PR #18, opened 2026-05-03
perf: reduce runtime CPU and memory footprint

EzgiTastan:
not an optimization engineer here, but I tried to collect the low hanging fruits

Measured locally on macos (:p), release build. Here is the comparison between baseline and this PR.

**Idle (sleep 90s):**
- CPU: 51% -> 6%
- RSS: 235 MB -> 208 MB

**Burst (20k lines output):**
- CPU: 47% -> 3%
- RSS: 237 MB -> 207 MB

**Binary:** 98 MB -> 76 MB

---

**What changed:**
- Switched bevy to explicit features instead of `default-features = true`
- Added `WinitSettings` with reactive low power mode
- Redraw throttle (16ms)
- Skip unnecessary work in Flat2d mode
- Reduce default scrollback (10000 -> 2000)
- Bounded PTY channel + larger read buffer

fyi, most of the diff is Cargo.lock (deleted unused bevy deps).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -110,28 +110,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
-name = "alsa"
-version = "0.9.1"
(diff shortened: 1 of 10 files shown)


orhun (approved):
aight

PR #25, opened 2026-05-11
feat: add Nix packaging, devshell, and modules

KristijanZic:
- Adds a flake.nix using crane and flake-parts for fast caching
- Provides a development shell with necessary dependencies
- Exports a Home Manager module to manage ratty.toml
- Exports a NixOS module to enforce system-wide config
- Patches hardcoded shell paths for NixOS compatibility

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,238 @@
+{
+  description = "A GPU-rendered terminal emulator that supports inline 3D graphics";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    flake-parts.url = "github:hercules-ci/flake-parts";
+    crane.url = "github:ipetkov/crane";
+  };
+
+  outputs =
+    inputs@{
+      self,
+      nixpkgs,
+      flake-parts,
+      crane,
+      ...
+    }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      systems = [
+        "x86_64-linux"
+        "aarch64-linux"
+        "x86_64-darwin"
+        "aarch64-darwin"
+      ];
+
+      perSystem =
+        {
+          config,
+          self',
+          inputs',
+          pkgs,
+          system,
+          lib,
+          ...
+        }:
+        let
+          craneLib = crane.mkLib pkgs;
+
+          buildInputs =
+            with pkgs;
+            [
+              vulkan-loader
+              libx11
+              libxcursor
+              libxi
+              libxrandr
+              wayland
+              libxkbcommon
+              fontconfig
+            ]
+            ++ lib.optionals stdenv.isDarwin [
+              darwin.apple_sdk.frameworks.Cocoa
+              darwin.apple_sdk.frameworks.AppKit
+              darwin.apple_sdk.frameworks.CoreGraphics
+              rustPlatform.bindgenHook
+            ];
+
+          nativeBuildInputs = with pkgs; [
+            pkg-config
+            makeWrapper
+            copyDesktopItems
+          ];
+
+          LD_LIBRARY_PATH = pkgs.lib.makeLibraryPath buildInputs;
+
+          src = craneLib.cleanCargoSource ./.;
+
+          cargoArtifacts = craneLib.buildDepsOnly {
+            inherit src buildInputs nativeBuildInputs;
+          };
+
+          ratty = craneLib.buildPackage {
+            inherit
+              src
+              cargoArtifacts
+              buildInputs
+              nativeBuildInputs
+              ;
+
+            desktopItems = [
+              (pkgs.makeDesktopItem {
+                name = "ratty";
+                desktopName = "Ratty";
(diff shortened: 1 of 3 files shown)


vimlinuz (commented):
lgtm!!!🚀

bubylou:
This built and ran on NixOS for me but for some reason the rat cursor was a white cube instead of the spinning rat.

flake.nix
```diff
@@ -0,0 +1,238 @@
...
+                comment = "A GPU-rendered terminal emulator with inline 3D graphics";
+                exec = "ratty";
+                terminal = false;
+                categories = [ "System" "TerminalEmulator" "Utility" ];
+                icon = "ratty";
+              })
+            ];
+
+            doCheck = false;
+            postInstall = pkgs.lib.optionalString pkgs.stdenv.isLinux ''
```
bubylou, line 93: Why are tests turned off, were they failing?

es-sai-fi:
what's the benefit of using flake-parts here?

servrox:
> This built and ran on NixOS for me but for some reason the rat cursor was a white cube instead of the spinning rat.

only got this for some custom models, since ratty falls back to the white cube when the cursor model path resolves to an absolute path outside its asset root. custom models only load reliably when placed under the expected `assets/objects` layout and referenced by a bare filename.

flake.nix
orhun: good question!

orhun:
Thanks for the PR! Any chance that we check the Nix flake in the CI as well?

flake.nix
```diff
@@ -0,0 +1,238 @@
...
+
+          LD_LIBRARY_PATH = pkgs.lib.makeLibraryPath buildInputs;
+
+          src = craneLib.cleanCargoSource ./.;
+
+          cargoArtifacts = craneLib.buildDepsOnly {
+            inherit src buildInputs nativeBuildInputs;
+          };
+
+          ratty = craneLib.buildPackage {
```
imnotpoz, line 72: please move this package to a separate file, you can later still use it in `inputsFrom` as `self.packages.${system}.ratty`

flake.nix
```diff
@@ -0,0 +1,238 @@
...
+            config = lib.mkIf cfg.enable {
+              home.packages = [ cfg.package ];
+
+              xdg.configFile."ratty/ratty.toml" = lib.mkIf (cfg.settings != { }) {
+                source = tomlFormat.generate "ratty.toml" cfg.settings;
+              };
+            };
+          };
+
+        nixosModules.default =
```
imnotpoz, line 170: same as the package with the nixos and home-manager modules, move them to separate files

I'd suggest `nix/package.nix`, `nix/nixos-module.nix` and `nix/home-manager-module.nix` (or move the modules into their own directory)

flake.nix
```diff
@@ -0,0 +1,238 @@
...
+      crane,
+      ...
+    }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      systems = [
+        "x86_64-linux"
+        "aarch64-linux"
+        "x86_64-darwin"
+        "aarch64-darwin"
+      ];
```
imnotpoz, line 24: maybe http://github.com/nix-systems/default as an input?

flake.nix
```diff
@@ -0,0 +1,238 @@
...
+
+  outputs =
+    inputs@{
+      self,
+      nixpkgs,
+      flake-parts,
+      crane,
+      ...
+    }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
```
imnotpoz, line 18: this could be removed, the attributes that don't depend on the system you could leave as is and the ones that do you can use the following function for:

```nix
forEachSystem = nixpkgs.lib.genAttrs (import systems);
```

where `systems` is the input mentioned in another comment

otherwise if you don't decide to use that input, replace `(import systems)` with the array of systems

imnotpoz (changes requested):

flake.nix
imnotpoz: I looked through the repo and didn't find any tests for ratty, which is why I also did `doCheck = false;` in the nixpkgs package

when nix builds a rust program it first builds it in release mode for the actual binary, then again in debug mode for tests (crane helps with it but you still need to build the ratty crate twice) - that slows down build time considerably and is the reason behind this

if I'm wrong about there being no tests please let me know, I'll update the package

KristijanZic:
> Thanks for the PR! Any chance that we check the Nix flake in the CI as well?

Yes, I can make a workflow.
I'll gather all the feedback from here and get to work. :)

----

Thanks everyone for the reviews. I really appreciate it!

Sporarum:
Maybe this deserves a different issue/PR, but we should also include nix to the install instructions and packaging status!

orhun:
now added in #83

Sporarum:
@orhun do you mean #61 ?

#83 seems to be a closed duplicate issue

orhun:
yup, meant that 👍🏼

PR #33, opened 2026-05-12
fix(rgp): reject traversal components and drop raw-path fallback

### LINKED ISSUE #32: RGP path traversal: terminal escape can load arbitrary files via cursor/object asset loader
## tl;dr

the RGP `r` (register) sequence takes a filesystem path from the terminal stream and loads it. `object_asset_path` tries to sanitize it, but `load_object_source` retries with the raw original path on the first error, which defeats the sanitizer. any process that can write to the terminal can make ratty read arbitrary `.obj` / `.glb` files on disk.

the classic threat: `cat poisoned.log` or a malicious ssh banner sends:

```
\x1b_ratty;g;r;id=1;fmt=obj;path=../../../../../tmp/x.obj\x1b\\
```

and ratty loads `/tmp/x.obj`.

## where

`src/model.rs:137-140`:

```rust
"obj" => load_obj_meshes_from_path(Path::new("assets").join(&candidate).as_path())
    .or_else(|_| load_obj_meshes_from_path(path))
    .map(|meshes| (candidate.clone(), ObjectSource::Obj(meshes))),
```

`candidate` is the *sanitized* path. `path` is the *original* attacker-controlled path. on any error opening the sanitized one (e.g. file not found because the sanitizer rewrote it), the loader retries with the raw original.

`object_asset_path` itself also doesn't filter `ParentDir` components when the path has no `assets` anchor (`src/model.rs:189-208`).

## PoC

standalone, no bevy. extracts `object_asset_path` verbatim and simulates the `Path::new("assets").join(...).or_else(...)` chain via `fs::canonicalize`. plant `/tmp/ratty-victim.obj`, run with various `path=` inputs:

```
attacker `path=` value: "../../tmp/ratty-victim.obj"
  candidate (sanitized): assets/../../tmp/ratty-victim.obj
  fallback (raw path)  : ../../tmp/ratty-victim.obj
    primary : does not resolve to an existing file
    fallback: opens /tmp/ratty-victim.obj (EXISTS)

attacker `path=` value: "../../../../../../../tmp/ratty-victim.obj"
  candidate (sanitized): assets/../../../../../../../tmp/ratty-victim.obj
  fallback (raw path)  : ../../../../../../../tmp/ratty-victim.obj
    primary : does not resolve to an existing file
    fallback: opens /tmp/ratty-victim.obj (EXISTS)
```

full PoC source: https://gist.github.com/bl4ckr0ss3/35afe3ed0a29a42345f1e8f2a67be371

## impact

1. reads any `.obj` / `.glb` file the user has access to. blind (no echo back over PTY), but:
   - confirms file existence via warn-log differences
   - causes the file to be parsed by `tobj` / bevy's gltf loader. any parser bug in those crates becomes reachable via a terminal escape sequence.
   - if the attacker can plant a file (world-writable dirs like `/tmp`, or a file the user just downloaded), the terminal will read it as a 3D mesh on their behalf. OBJ supports `mtllib` references which the on-disk loader follows, so the attacker can chain to a second file.

2. this is the same threat class as the iTerm2 / xterm escape-sequence CVEs (CVE-2017-7494 etc). standard advice for terminals: never let an APC sequence cause filesystem side-effects with an attacker-supplied path.

reachable from anywhere that writes to the PTY: `cat`, `less`, SSH MOTD, shell prompts that print remote content, etc.

## suggested fix

the minimal fix is two changes:

1. drop the `.or_else(|_| load_obj_meshes_from_path(path))` fallback. always go through the sanitized path.
2. in `object_asset_path`, filter `ParentDir` / `RootDir` / `CurDir` components in all branches, not just the post-`assets` one.

plus: after building the resolved path, canonicalize it and verify it lives under the project's `assets/` root. that gives a second layer of defense in case the embedded `RustEmbed` lookup or the gltf branch goes wrong.

happy to send a PR with the fix once you confirm the direction. also have a few smaller things in the same file (APC parser memory DoS via incomplete sequences, PNG decompression bomb in `kitty.rs` finalize, kitty multipart bytes unbounded growth) but those are separate issues, want to keep this one focused.

## env

ratty `1533591` (main as of 2026-05-12), debian/kali, x86_64.

bl4ckr0ss3:
closes #32

## what

two changes in `src/model.rs`:

1. drop the `.or_else(|_| load_obj_meshes_from_path(path))` fallback in `load_object_source`. it retried with the raw, unsanitized attacker path on any error from the sanitized lookup, which defeated `object_asset_path` for any input the sanitizer rewrote into a non-existent location.
2. reject `ParentDir` / `RootDir` / `CurDir` components in both branches of `object_asset_path` (the `assets`-anchor branch and the no-anchor fallback). legitimate inputs are unchanged.

## verified

added five unit tests in `src/model.rs::tests` covering:

- relative `../` traversal -> rejected
- traversal after an `assets/` anchor -> rejected
- absolute path -> rejected
- bare filename -> still resolves to `objects/<file>`
- `assets/objects/<file>` -> still resolves to `objects/<file>`

also ran a standalone repro (no bevy/tobj, plain rustc) that mirrors the full `load_object_source` resolution chain against both attack and legitimate inputs - all attacks now rejected, all legitimate paths unchanged. output:

```
[ok]   "../../tmp/x.obj" -> rejected (asset path contains traversal components: ../../tmp/x.obj)
[ok]   "../../../../../../../tmp/ratty-victim.obj" -> rejected (...)
[ok]   "assets/../etc/passwd.obj" -> rejected (...)
[ok]   "/etc/passwd.obj" -> rejected (absolute path is outside the asset root)
[ok]   "CairoSpinyMouse.obj" -> "objects/CairoSpinyMouse.obj"
[ok]   "assets/objects/CairoSpinyMouse.obj" -> "objects/CairoSpinyMouse.obj"
[ok]   "assets/foo.obj" -> "foo.obj"
[ok]   "subdir/foo.obj" -> "subdir/foo.obj"
```

standalone repro source: https://gist.github.com/bl4ckr0ss3/35afe3ed0a29a42345f1e8f2a67be371 (i'll push the fixed version there too if useful).

## not in this PR

there are a few smaller adjacent things i noticed while auditing (APC parser memory DoS via incomplete sequences in `inline.rs`, kitty PNG decode without size limits in `kitty.rs::finalize`, kitty multipart byte buffer unbounded growth). i'll open separate issues for those, keeping this PR focused on the traversal fix.


orhun:
Thanks for the detailed issue and the PR!

This looks good mostly, but I wasn't able to run the widget examples. Especially `document` and `draw`:

```
2026-05-12T15:56:28.693336Z  WARN ratty::inline: failed to load RGP object 100: failed to read assets/sprite_12_offset_32218.obj: open file failed
2026-05-12T15:56:28.693658Z  WARN ratty::inline: failed to load RGP object 101: failed to read assets/black.obj: open file failed
2026-05-12T15:56:30.151345Z  WARN ratty::inline: failed to load RGP object 102: failed to read assets/bomber.obj: open file failed
2026-05-12T15:56:30.151383Z  WARN ratty::inline: failed to load RGP object 103: failed to read assets/battle.obj: open file failed
2026-05-12T15:56:30.334139Z  WARN ratty::inline: failed to load RGP object 104: failed to read assets/sprite_18_offset_48048.obj: open file failed
2026-05-12T15:56:31.465527Z  WARN ratty::inline: failed to load RGP object 102: failed to read assets/bomber.obj: open file failed
2026-05-12T15:56:31.465582Z  WARN ratty::inline: failed to load RGP object 103: failed to read assets/battle.obj: open file failed
2026-05-12T15:56:31.648197Z  WARN ratty::inline: failed to load RGP object 101: failed to read assets/black.obj: open file failed
2026-05-12T15:56:31.782133Z  WARN ratty::inline: failed to load RGP object 100: failed to read assets/sprite_12_offset_32218.obj:
```

Can you adjust the widget examples based on the new asset loading changes as well?

bl4ckr0ss3:
thanks for testing! pushed a follow-up commit that updates the widget examples to work with the new restricted loader.

changes:

- moved `widget/assets/*.obj` -> `assets/widget/*.obj` so they live under the project asset root
- `document` example sends `widget/<name>` (relative, sanitizer-friendly) instead of the previous absolute workspace path. ratty resolves that to `assets/widget/<name>` via the same logic that handles bare filenames.
- `draw` writes its live preview to `assets/widget/.live_draw.obj` (gitignored) and sends `widget/.live_draw.obj`
- `TempleOS.jpg` stays at `widget/assets/TempleOS.jpg` since the widget reads it directly via `image::ImageReader::open` (ratty doesn't see it)

verified `cargo build --examples` in `widget/` is clean. the unit tests in `model.rs::tests` still pass against the original sanitizer changes.

lmk if you want big_rat updated too for consistency (it sends an absolute path to `assets/objects/SpinyMouse.glb` which currently works because the sanitizer strips up to the `assets` component and that file already lives at `assets/objects/`), or if it's fine as-is.

orhun:
Moving widget assets to the root assets folder is not a good solution. It means that we will start embedding those assets into the Ratty binary which is unnecessary.

bl4ckr0ss3:
rust_embed only globs ```assets/objects/``` (verified in ```src/model.rs:15-17```), so I don't think ```assets/widget/```  would end up at blob, but I understand the concern so I'll look into the other way but first I need to cook food for the dinner >..<

orhun:
oh got it, you're right!

orhun:
but then those assets won't be a part of the widget's crates.io release (which only includes the widget directory) which is still problematic :/

I think we should have those widgets assets in the widget folder respectively

orhun:
any updates on this one?

PR #36, opened 2026-05-12
fix: fall back to $SHELL or /bin/sh by default

### LINKED ISSUE #27: Ratty fails to spawn shell on systems without /bin/bash
## Summary

  Ratty crashes on first launch with `Error: failed to spawn shell — Unable to spawn /bin/bash (ENOENT)` on systems that don't ship `/bin/bash` (NixOS, some BSDs, minimal
  containers). The default config hardcodes `[shell] program = "/bin/bash"`, which is checked before the existing `$SHELL` fallback in `src/runtime.rs`, so the fallback never runs.

  ## Steps to reproduce

  1. Install ratty on a system with no `/bin/bash` (e.g. NixOS).
  2. Run `ratty` with no user config (uses the bundled `config/ratty.toml`).
  3. Crash.

  ## Expected behavior

  Ratty should launch the user's `$SHELL`, falling back to `/bin/sh` (POSIX-mandated and universally present) only if `$SHELL` is unset.

  ## Environment

  - OS: NixOS (kernel 6.18.26)
  - Ratty version / commit: 0.2.0 (`8b14267`)
  - GPU / graphics stack: (irrelevant — fails before any rendering)
  - Config changes: none — reproduces with default `config/ratty.toml`

  ## Additional context

  When I was goofing with nix builds, I got some reference error from `nix run`:
```bash
  Error: failed to spawn shell
```
  Caused by:
      Unable to spawn /bin/bash because it doesn't exist on the filesystem or is not executable (ENOENT: No such file or directory)

  Source of the issue: `config/ratty.toml` sets `program = "/bin/bash"`, and `src/runtime.rs:215-221` checks `config.shell.program` before `env::var("SHELL")`, so the explicit
  config value always wins.

  Fix is two lines (PR ready on `fix/shell-default`):
  - Remove `program = "/bin/bash"` from `config/ratty.toml`.
  - Change `/bin/bash` → `/bin/sh` for the final fallback in `runtime.rs:221`.

orhun:
fixes #27

### DIFF
--- a/config/ratty.toml
+++ b/config/ratty.toml
@@ -9,9 +9,9 @@ default_cols = 104
 default_rows = 32
 scrollback = 2000
 
-[shell]
-program = "/bin/bash"
-args = []
+# [shell]
+# program = "/bin/bash" 
+# args = []
 
 [env]
 TERM = "xterm-256color"
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -218,7 +218,7 @@ impl TerminalRuntime {
                 .as_ref()
                 .map(|path| path.to_string_lossy().into_owned())
                 .or_else(|| env::var("SHELL").ok())
-                .unwrap_or_else(|| "/bin/bash".to_string());
+                .unwrap_or_else(|| "/bin/sh".to_string());
             let mut cmd = CommandBuilder::new(shell);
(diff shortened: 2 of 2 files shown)


TheCodedKid:
Cool, glad the fix worked.

PR #38, opened 2026-05-13
Updated README.md to run the install command for arch and dependency install commands for other Distros

zalanwastaken:
# Improve installation instructions across distributions

### Updated the README.md to provide clearer and more accurate installation steps.

* Updated  install command for Arch Linux users
* Updated dependency installation instructions for other Linux distributions


orhun (approved):
thx!

PR #62, opened 2026-05-17
fix: enable shading for all objects

qewer33:
This fix enables shading for all objects instead of just objects with a non-zero `depth` property. The `use_lighting` variable is just always set to true now, I didn't remove it as in the future it could be used as is for adding a shading enable/disable toggle to the protocol.


orhun (approved):
thx!!

Issue #15, opened 2026-04-30
pls make this cross platform


programmersd21:
weee

orhun:
ok what platform

btw it works on macos

https://bsky.app/profile/vimlena.com/post/3mkoshbzpvs2y

https://github.com/user-attachments/assets/b4b662ab-286d-4c35-899d-b5b11c3cfe97

programmersd21:
need windows support too

orhun:
The hard part of making it support Windows is not the 3D rendering but the terminal runtime itself: currently we depend on [portable-pty](https://docs.rs/portable-pty) for the PTY behavior and the rest of the terminal stack is built under Linux-like assumptions. So Windows support would require real platform-specific work and validation.

PR #68, opened 2026-05-18
feat: support Windows via OSC envelope

### LINKED ISSUE #67: RGP examples don't render inline 3D on Windows — ConPTY strips APC sequences
## Summary

Ratatui-ratty examples (`big_rat`, `draw`, `document`) and the `ratatui-ratty` widget in general do not render inline 3D objects when run inside Ratty on Windows. The terminal opens, the TUI header / canvas renders correctly, but the 3D preview area stays empty and no `registered RGP object …` / `failed to load RGP object …` line is ever logged.

Splitting this off from #66 (Windows shell defaults) per discussion there — that PR is the right way to fix #44, this is a separate, deeper issue.

## Root cause

ConPTY strips APC sequences before they reach the host process. Ratty's RGP register/place/update sequences (which all use APC: `\x1b_ratty;…\x1b\`) never arrive at `inline::consume_pty_output`, so `handle_rgp_sequence` is never called and no object is registered.

### Reproduction without Ratty

A minimal `portable-pty`-backed probe that runs `bash -c "printf '\x1b_PROBE;hello=world\x1b\\'"` confirms it:

```
=== PROBE RESULT (default ConPTY flags) ===
Total bytes read: 16
0000: 1b 5b 3f 39 30 30 31 68 1b 5b 3f 31 30 30 34 68    .[?9001h.[?1004h

Contains \x1b_ (APC start)?  false
Contains 'PROBE;' marker?    false
```

Only the bash-init CSIs survive — the APC body and ST terminator are dropped by conhost.

### Reproduction inside Ratty

`ratty.exe -e <path>\big_rat.exe` from inside the ratty repo. With `RUST_LOG=warn`:

- `loaded cursor model from embedded:CairoSpinyMouse.obj (1 mesh parts)` — fires (cursor model registered via the in-process path, no PTY involved).
- `registered RGP object …` — **never fires**.
- `failed to load RGP object …` — **never fires**.

No write happens to `assets/objects/external/`, confirming `ensure_scene_asset_path` is never reached for the example's GLB. So the issue is not asset materialization, path separators, or Bevy GLTF loading — it's that the APC stream never reaches Ratty at all.

## Fix path

Microsoft added `PSEUDOCONSOLE_PASSTHROUGH_MODE` (`0x8`) for exactly this use case in Windows 11 22H2+ / Windows Server 2025. With that flag, conhost forwards APC/OSC/DCS sequences verbatim to the host instead of consuming them itself.

`portable-pty` already declares the constant:

```rust
#[allow(dead_code)]
pub const PSEUDOCONSOLE_PASSTHROUGH_MODE: DWORD = 0x8;
```

but never OR's it into the `CreatePseudoConsole` call. Until that's exposed, Ratty (and any other terminal built on `portable-pty`) can't see APC traffic on Windows regardless of what they do on the parsing side.

### Suggested approach

1. Upstream PR against `wezterm/wezterm` (the `portable-pty` crate) adding an env-var (`PORTABLE_PTY_CONPTY_PASSTHROUGH=1`) or builder-API opt-in for the flag. Default off so existing callers are unchanged.
2. Once released, bump `portable-pty` in Ratty's `Cargo.toml`, and either:
   - default the flag on inside Ratty's runtime, or
   - expose it via `ratty.toml` (e.g. `[terminal] conpty_passthrough = true`).

Happy to drive both PRs — flagging here so it can be tracked.

## Environment

- Windows 11 Pro 26200.6457
- NVIDIA GeForce RTX 5090 (Vulkan)
- Ratty 0.3.0 + #66 (Windows shell defaults)
- `portable-pty` 0.8.1 (current Ratty pin)

ahostbr:
## Summary

Adds OSC envelope (`ESC ] 8901 ; ratty ; g ; … ESC \`) as an alternative to the existing APC envelope for the Ratty Graphics Protocol, and fixes a related scroll-tracking false-positive. Together these unblock RGP-based widgets — including the bundled `big_rat`, `draw`, and `document` examples — on Windows. Fixes #67.

## Why

Windows ConPTY strips APC sequences (`ESC _ … ESC \`) before they reach the host process. Verified with a minimal `portable-pty`-backed probe: a child running `printf '\x1b_PROBE;hello=world\x1b\'` arrives at the PTY master with only the bash init CSIs (`\x1b[?9001h\x1b[?1004h`) — APC body and ST terminator gone. Equivalent OSC payloads at arbitrary numbers (1337, 9001, 99999, 52, 0) survive intact end-to-end.

This is structural in conhost, not a configuration issue:

- `PSEUDOCONSOLE_PASSTHROUGH_MODE` isn't a real public flag. The value `0x8` in current `winconpty.h` is `PSEUDOCONSOLE_GLYPH_WIDTH_GRAPHEMES` (not passthrough). The `#[allow(dead_code)]` constant in `portable-pty` traces back to a 2019 proposal that never shipped.
- Side-loading the latest `microsoft/terminal` v1.24 `conpty.dll` + `OpenConsole.exe` does not change the APC behavior (empirically tested). The 1.22 "passthrough" PR ([#17510](https://github.com/microsoft/terminal/pull/17510)) was an internal conhost refactor that enables conhost to render unrecognised sequences (Sixel) in its own renderer — not a forwarding feature for third-party hosts.
- OSC, by contrast, is preserved by conhost for any number it doesn't have a hardcoded handler for. Issue #67 has the full investigation.

## Changes

**`src/rgp.rs`** — adds `RGP_OSC_START = b"\x1b]8901;ratty;g;"`. `consume_sequence` is refactored to recognise either envelope and strip the right terminator (APC ends on ST/C1-ST; OSC also accepts BEL). The body grammar (`<verb>;<key=value>;…`) is unchanged, so APC and OSC paths share the same parser.

**`src/inline.rs`** — `consume_pty_output`'s scan loop gains an `EnvelopeKind` discriminator; it picks the next APC (`\x1b_`) *or* the specific RGP OSC prefix and dispatches each through the right path. Unrelated OSC traffic (window titles, hyperlinks, prompt marks) flows straight through to vt100 as before.

**`src/inline.rs` + `src/systems.rs`** — fixes a pre-existing scroll-tracking false-positive that the OSC migration exposes. When a `Place` arrives in the same PTY chunk that clears and repaints the terminal, `infer_upward_scroll` can return a large shift (25 rows on a 32-row screen in this repro) and the old `apply_scroll` evicts the very anchor that chunk just set. Tracked via a new `placed_this_chunk: HashSet<u32>` on `TerminalInlineObjects`; `pump_pty_output` drains it after consume and scrolls only `pre_anchor_ids \ placed_this_chunk`. Public `apply_scroll(rows)` API is unchanged; new `apply_scroll_to_existing(rows, exempt)` is the underlying helper.

**`widget/src/lib.rs`** — `ratatui-ratty`'s seven emit sites (register path, register payload init/chunk/empty, place, update, delete) switch from APC to OSC 8901. ST terminator emitted; parser also accepts BEL.

## OSC 8901 — number choice

Picked as a 4-digit private number with no known assignment in any surveyed terminal codebase, registry, or shell-integration spec. Numbers explicitly avoided:

| Number | Used by |
|---|---|
| 0,1,2,4,7,8,9,10,11,12,52,104,110-119 | conhost hardcoded |
| 99 | kitty notifications |
| 133 | FinalTerm / Warp prompt marks |
| 633 | VS Code shell integration |
| 777 | urxvt |
| 1337 | iTerm2 image protocol |
| **9001** | **Microsoft Terminal's `WTAction`** in `src/terminal/parser/OutputStateMachineEngine.hpp` — dispatching is numeric-first, so the `;ratty;` discriminator does not protect against a numeric collision |

The `;ratty;` discriminator after the number namespaces the payload for the protocol family, leaving room for `g;` (graphics) plus future verb-spaces if you want them.

## Verification

Windows 11 Pro 26200.6457 / RTX 5090 / Vulkan backend. `cargo fmt --check`, `cargo check`, `cargo clippy --all-targets --all-features -- -D warnings` all green.

**Wire-level probe** (vendored `portable-pty 0.8.1` shelling out to `bash.exe`):

```
=== APC (before) ===
Total bytes read: 16
0000: 1b 5b 3f 39 30 30 31 68 1b 5b 3f 31 30 30 34 68    .[?9001h.[?1004h
Contains \x1b_ (APC start)?  false
Contains 'PROBE;' marker?    false

=== OSC 8901 (after) ===
Total bytes read: 167
0000: 1b 5b 3f 39 30 30 31 68 1b 5b 3f 31 30 30 34 68 1b 5b 3f 32 35 6c …
0040: …  .]9001;r…[trimmed]
0080: …;animate=1;scale=1.0;depth=0.0;color=ffffff;brightness=1.0;…
Contains \x1b]8901;ratty;g;…?  true
```

**End-to-end inside `ratty.exe`** — `ratty -e big_rat.exe`:

```
INFO ratty::inline: registered RGP object 7 from assets/objects/SpinyMouse.glb
```

The full RGP register → place → spawn → render chain fires, and the SpinyMouse GLB renders inside Ratty's viewport for the first time on Windows. Screenshot below.

image

## Compatibility

Existing APC senders keep working. The parser tries both envelopes per sequence, and `ratatui-ratty`'s on-the-wire behavior is the only change for downstream widget consumers — they get a one-time API recompile but nothing breaks in the body grammar.

## Open follow-ups (separate work)

- ChatGPT deep research flagged that mid-line OSC emission can interact with a conhost ordering bug ([microsoft/terminal#17313](https://github.com/microsoft/terminal/issues/17313)) on pre-1.23 conhost. RGP `place` is already self-describing (`row`, `col`, `w`, `h` carry absolute coordinates), so reordering does not silently mis-anchor the object, but a freshness key (`seq=` or `frame=`) would defeat stale-arrival cases. Worth a follow-up if it ever bites in practice.
- Multiplexers (tmux, screen, zellij) intercept selected OSC numbers and may need a `\ePtmux;…\e\` passthrough wrapper if a user runs RGP-emitting apps through them. Out of scope here.

### DIFF
--- a/protocols/graphics.md
+++ b/protocols/graphics.md
@@ -19,6 +19,13 @@ but it is designed for Ratty's inline object layer and 3D renderer.
 
 ## Transport
 
+RGP sequences are framed by a standard terminal escape envelope. Two
+envelopes are defined — **APC** (the primary) and **OSC** (the
+compatibility alternative). Both carry identical body grammar; only the
+framing bytes differ.
+
+### APC Envelope (primary)
+
 Ratty Graphics Protocol uses [APC] (Application Program Command):
 
 ```text
@@ -31,6 +38,58 @@ Where:
 - `g` means graphics
 - [`<verb>`](#verbs) selects the operation
 - additional fields are semicolon-separated `key=value` pairs
+
+### OSC Envelope (compatibility)
+
+On platforms where the parent terminal cannot transit APC sequences
+end-to-end, the same body can be carried inside an [OSC] (Operating
+System Command) envelope using private number **8901**:
+
+```text
+ESC ] 8901;ratty;g;<verb>[;<key=value>...] ESC \
+```
+
+The body after `8901;` is byte-identical to the APC body (starting from
+`ratty;g;`). The OSC envelope additionally accepts **BEL** (`0x07`) as
+an alternative terminator, following the historic xterm convention.
+
+#### When to use OSC
+
+The OSC envelope exists for one reason: **Windows ConPTY** silently
+consumes APC sequences before the hosting terminal can see them. This
+is true even with the newer `PSEUDOCONSOLE_INHERIT_CURSOR` plumbing in
+[Microsoft Terminal] 1.22+. Applications that may run under ConPTY
+(cmd, PowerShell, Windows Terminal, SSH into Windows) should emit the
+OSC envelope so that Ratty can parse the sequences.
+
+Applications running on platforms without this limitation (direct PTY on
+Linux/macOS, or inside Ratty's own PTY) should prefer the APC envelope.
+
+#### Why OSC 8901
+
+OSC 8901 was chosen as a 4-digit private number with no known assignment
+in any surveyed terminal ecosystem. Numbers to avoid:
+
+| Number(s) | Owner |
+|-----------|-------|
+| 0–12, 52, 104, 110–119 | conhost / xterm standard |
+| 99 | Kitty notifications |
+| 133 | FinalTerm / Warp prompt marks |
+| 633 | VS Code shell integration |
+| 777 | urxvt |
+| 1337 | iTerm2 image protocol |
+| 9001 | `WTAction` in Microsoft Terminal's `OscActionCodes` enum |
+
+OSC dispatching in conhost is numeric-first — the `;ratty;` qualifier
+after the number does not prevent a terminal from consuming the whole
+sequence as its own action. A unique number avoids collisions entirely.
+
+#### Support reply
+
+The support reply (`s` verb) is always sent using the APC envelope,
+regardless of which envelope the query arrived in, because the reply
+travels from Ratty to the application over the PTY read path where APC
+is safe.
 
 ## Model
 
--- a/src/inline.rs
+++ b/src/inline.rs
@@ -10,12 +10,24 @@ use vt100::Callbacks;
 use crate::kitty::{KittyOperation, KittyParserState, refresh_kitty_placeholder_anchors};
 use crate::model::{ObjectSource, load_object_source, load_object_source_from_bytes};
 use crate::rgp::{
-    RgpOperation, RgpPlacementStyle, RgpPlacementUpdate, RgpRegisterSource,
+    RGP_OSC_START, RgpOperation, RgpPlacementStyle, RgpPlacementUpdate, RgpRegisterSource,
     consume_sequence as consume_rgp_sequence, support_reply,
 };
 const APC_START: &[u8] = b"\x1b_";
 const ST: &[u8] = b"\x1b\\";
 const C1_ST: u8 = 0x9c;
+const BEL: u8 = 0x07;
+
+/// Envelope variants we slice out of the PTY byte stream for RGP/Kitty
+/// dispatch. APC catches any APC sequence (Kitty graphics, RGP); the OSC
+/// variant is specifically pre-filtered to our RGP prefix so unrelated OSC
+/// traffic (window titles, hyperlinks, prompt marks) passes straight to
+/// vt100.
+#[derive(Clone, Copy)]
+enum EnvelopeKind {
+    Apc,
+    RgpOsc,
+}
 
 /// Marker for 2D inline object sprites.
 #[derive(Component)]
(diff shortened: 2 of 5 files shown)


orhun:
Does it require any updates in the protocol document?

ahostbr:
Good call — pushed `a29fabf` which adds a Transport subsection to `protocols/graphics.md` covering:

- **OSC 8901** as a compatibility envelope alongside the existing APC envelope
- When to use it (Windows ConPTY strips APC silently)
- The OSC number selection rationale with a collision-avoidance table (conhost, Kitty, iTerm2, VS Code, WTAction 9001, etc.)
- That both envelopes carry identical body grammar — only framing differs
- Support reply always uses APC regardless of query envelope (safe direction: Ratty → app)

orhun:
Thanks for clarifying. I probably need to test this on Windows before I go ahead and review it.
I currently don't have a Windows setup, do you know if it is possible to test it out on Linux somehow?

I also appreciate more eyes on this PR  🙏🏼 If anyone uses Windows and tries Ratty, feel free to chime in on this.

ahostbr:
you could run a windows emulator on linux to test it thats probably the cleanest route.  I confirm it runs on my machine and windows 11 sandbox.

programmersd21:
please merge this @orhun would help in windows support as i mentioned in #15

orhun:
I haven't tested this and don't want to yolo it without seeing it in action on windows :/

orhun:
Also there seems to be some conflicts (and more coming with #107) - it probably needs some conflict resolution at some point.

PR #76, opened 2026-05-22
Revert font fallback options

gold-silver-copper:
## Summary

Reverts the added terminal renderer font fallback families and restores the renderer to use only the configured font family.

## Why

This undoes the behavior introduced by PR #55, removing the hard-coded fallback family list from `build_terminal_renderer`.

## Impact

Users return to the previous font selection behavior based on the configured font family and size.

## Validation

- `cargo test`

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -312,14 +312,7 @@ fn build_terminal_renderer(
         ),
         palette,
     };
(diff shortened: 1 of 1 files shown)


orhun:
I think I'm gonna go with this either way

orhun (approved):

PR #79, opened 2026-05-26
docs(readme): add comchan to list of apps

Vaishnav-Sabari-Girish:
This adds [ComChan](https://github.com/Vaishnav-Sabari-Girish/ComChan) to the list of apps using `ratty`


orhun (approved):
thanks!

PR #86, opened 2026-05-30
fix: Make xy plane offset coordinates in 2D mode work like in 3D

### LINKED ISSUE #85: Translation offsets in 2D mode and 3D mode are completely different
It seems the px and py offsets on the update command of the graphics protocol are in normalized device coordinates format in 3D mode (offset of 1.0 = a whole screen), while in some other scale in 2D mode (offset of 1.0 = barely noticeable).

It may very well be that I'm doing something wrong, test code for the demo video [here](https://github.com/Nomagno/nomrat/blob/58fc26a422b73652a8ad3365c24195e302cf2b5e/test.c) with the single header wrapper for the Ratty Graphics Protocol [here](https://github.com/Nomagno/nomrat/blob/58fc26a422b73652a8ad3365c24195e302cf2b5e/nomrat.h).

In the attached video, the flower almost isn't moving in 2D mode because I am targetting the 3D mode scale.

https://github.com/user-attachments/assets/bac5d595-3b2f-40ae-8741-d5001e1dbbc1

Nomagno:
Fix #85
(It's the first of the two proposed solutions)

### DIFF
--- a/src/systems.rs
+++ b/src/systems.rs
@@ -890,8 +890,8 @@ pub(crate) fn sync_rgp_objects(mut params: RgpSyncParams) {
         match presentation.mode {
             TerminalPresentationMode::Flat2d => {
                 transform.translation = Vec3::new(
-                    layout.center_x + anchor.style.offset.x,
-                    layout.center_y + bob + anchor.style.offset.y,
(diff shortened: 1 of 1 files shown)


orhun (approved):
thanks!

PR #87, opened 2026-05-30
feat(windows): embed app icon and set consoleAllocationPolicy=detached

sitiom:
- Set consoleAllocationPolicy=detached on Windows to only show console output when inside an existing console session
- Embed and set the application icon

https://github.com/user-attachments/assets/d23511d0-29c5-44cb-afa9-d34fee1f27ed


src/main.rs
```diff
@@ -66,8 +79,47 @@ fn main() -> anyhow::Result<()> {
...
                 }),
         )
+        .add_systems(Update, apply_window_icon)
         .add_plugins(TerminalPlugin)
         .run();
 
     Ok(())
 }
+
+fn apply_window_icon(
```
orhun, line 95: Can you add a one-line comment on what this function is doing

src/main.rs
```diff
@@ -66,8 +79,47 @@ fn main() -> anyhow::Result<()> {
...
+
+            if !primary_window.visible {
+                window.set_visible(true);
+                primary_window.visible = true;
+            }
+        });
+    }
+}
+
+fn load_window_icon() -> anyhow::Result<Icon> {
```
orhun: How often do we call this during the Bevy app's lifecycle?

I'm thinking whether if we could do something else than encoding the bytes on memory.

Also, can you document this function?

build.rs
orhun, line 1: Can you explain what's going on here? I'm not familiar with the windows specific window handling

src/main.rs
sitiom: > How often do we call this during the Bevy app's lifecycle?

Once during startup. The `WindowCreated` system only reuses the already-created icon.

> I'm thinking whether if we could do something else than encoding the bytes on memory.

Good point. `build.rs` already embeds the ICO as a Windows resource, so I'll use that instead for Windows. I'll keep the previous logic for X11.

build.rs
sitiom: `winresource` embeds the app icon and application manifest into the executable.

The manifest uses `consoleAllocationPolicy=detached`, which keeps the binary as a console app when launched from an existing terminal, but avoids opening an extra console window when it is launched from Explorer/Start.

orhun (approved):
LGTM, thanks!

PR #89, opened 2026-05-31
Add perspective projection, ability to hotswap between orthographic and perspective

Nomagno:
Hey!
Ctrl + Alt + O -> Regular orthographic view
Ctrl + Alt + P -> New perspective view ('videogame camera' style)
Basically, both cameras are created at startup, but only the requested one is activated in Bevy. Not sure if this might suppose a performance penalty. I am also not an expert in Rust nor Bevy, so sorry if something's improvable :sweat_smile:. I also refactored a bunch of `match` statements into comparisons against `is_3d()` where I thought it might be reasonable, since 3 long enum values per arm seemed silly once I added the new `Persp3d` mode (I can revert this if needed).

I can make an issue if needed to discuss this further, just thought it was worth making a PR over since I had already implemented it. Demo video attached with the fixes from my other PR (for demo purposes since it uses the animation system).

https://github.com/user-attachments/assets/a180a2ce-8781-4d2e-ad88-77863fa76726

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -202,9 +202,12 @@ pub enum BindingAction {
     /// Disables a binding.
     #[serde(rename = "none")]
     None,
-    /// Toggles between the flat and warped terminal views.
-    #[serde(rename = "Toggle3DMode")]
-    Toggle3DMode,
+    /// Toggles between the flat and orthographic terminal views.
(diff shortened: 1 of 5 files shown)


arf20:
nice

Nomagno:
Obsoleted by #118

PR #90, opened 2026-06-02
Experimental cursor missile effect

TanTanDev:
### (experimental, draft pull request)
I've implemented a cursor effect:
ratty-f002726

[YouTube video showcasing the effect](https://youtu.be/agoO5lx77FY?si=tGqmUWP-0pZa6nb6)

### What it does
When the cursor is detected to have moved, an entity with (3D model) OR (texture) spawn where the cursor was. this "missile" entity flies towards the new cursor location, and when it arrives, it spawns a "hit" texture. You can assign spawn and hit sound effects.

I implemented 2 versions, runescape and maplestory. All sound / art i added was made by me by hand.

### Disclaimer
This is a draft pull request, because it's not fully implemented, and I don't intend to spend more time on this.
But since it's such a cool feature, I would like to share my implementation, if anyone want to pick it up.

### What it would take to integrate
The implementation I made luckily was able to be put into a single plugin 'CursorEffectPlugin'.
The moment it's added to the bevy App, the cursor effect will always happen. There is no way right now to configure OR DISABLE the cursor, except via the code.

**Known Issues:**
* Cursor effect, triggers when rotating in the 3D view
* no serialization configuration
* no sound settings (always max volume, no pitch variance)
* hit-texture Z order is messed up in 3D view. (it appears behind the plane)

**SIDE-EFFECTS: Asset loading**
When I ran the app with cargo r, bevy didn't look for asset folder from my repo, instead it looked inside /home/user/.cache/ratty/assets. to force the app to load from "assets" relative to the environment directory, I did this code change, which solved the issue for me. This should be examined for side-effects.
```rs
pub fn runtime_asset_root() -> PathBuf {
    if let Ok(cwd) = std::env::current_dir() {
        let local_assets = cwd.join("assets");
        if local_assets.is_dir() {
            return local_assets;
        }
    }
    ... the rest of the original code ...
```
**code change disclaimer**
I had to make some functions public from systems.rs like active-mobious_progress, cursor_pose, and a bunch of structs inside systems.rs. Some of these functions I marked  #allow(missing_docs)]

**crate changes**
* I added bevy feature "vorbis" to support playing .ogg files.
* I added "rand" to randomize the initial velocity of the missile.


v4u6h4n:
I hope I'm speaking for everyone when I say this feature is exactly what ratty was created for.

MordechaiHadad:
Incredible addition, should be considered @orhun

PR #96, opened 2026-06-04
Fix: report correct window size back to Portable-PTY at all times

### LINKED ISSUE #95: Window size not reported correctly
Test code from https://sw.kovidgoyal.net/kitty/graphics-protocol/#getting-the-window-size

The following code:
```c
#include <stdio.h>
#include <sys/ioctl.h>

int main(int argc, char **argv) {
    struct winsize sz;
    ioctl(0, TIOCGWINSZ, &sz);
    printf(
        "number of rows: %i, number of columns: %i, screen width: %i, screen height: %i\n",
        sz.ws_row, sz.ws_col, sz.ws_xpixel, sz.ws_ypixel);
    return 0;
}
```
Produces the following output on Ratty, regardless of the actual window resolution (the terminal rows/columns do vary properly):
```
number of rows: 31, number of columns: 96, screen width: 0, screen height: 0
```

Nomagno:
Fixes #95
The reported resolution is that of 2D mode, as that's the viewport size and it is also used by the Kitty Graphics Protocol and such, and by the animation system's normalized device coordinates.

### DIFF
--- a/src/keyboard.rs
+++ b/src/keyboard.rs
@@ -483,7 +483,7 @@ pub fn handle_keyboard_input(
                             ((params.viewport.size.x / char_dims.x as f32).floor() as u16).max(1);
                         let rows =
                             ((params.viewport.size.y / char_dims.y as f32).floor() as u16).max(1);
-                        params.runtime.resize(cols, rows);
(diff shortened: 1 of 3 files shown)


orhun (approved):
LGTM, thanks!

Issue #78, opened 2026-05-24
Texture / material support


Doridian:
## Summary

I would like a way for ratty to not just load 3D models, but also textures for them

## Motivation

Right now, models seem to have the ability to have colors, and that seems to be the limit of styling them.

I would like to see if more complex usages are possible, mainly for two directions:

- The cursor right now is a single 3D object, so it limits the amount of "fun" you can have with it. I'd love a rotating 3D fox, but in single-color, that does not achieve what I'm going for.
- I would like to investigate silly projects, like a renderer for Minecraft maps that works in the terminal (such that I can browse world files from the terminal and such)

## Proposed approach

The used OBJ and GLB formats already have accompanying files (MTL and GLTF iirc), which should be able to be loaded/streamed similarly.

I might investigate this myself once I'm back from traveling, I just wanted to get this idea written down in case anyone has opinions or an approach already

orhun:
Hey, thanks for the issue!

I totally agree with this and we should definitely support loading texture files alongside of models.
I'm not very familiar with the mentioned formats so I should ask, isn't GLB files already support colors etc. so you could in theory load an object with multiple colors?

We could probably implement loading MTL files along with OBJ files. I'm not sure if GLTF is an accompanying format to GLB or a format on its own.

Doridian:
> We could probably implement loading MTL files along with OBJ files. I'm not sure if GLTF is an accompanying format to GLB or a format on its own.

I am not sure about GLB at all. I've only learnt about that format thanks to ratty. So far I've always been in the OBJ or STL world of 3D model formats (and OBJ has been a while)

orhun:
Got it. I think GLTF is a file format on its own and it might be nice to support it as well.

I mostly use objects from https://poly.pizza and it supports both GLB and OBJ. That's why I went them in the first place.

A good first step at implementing this would be to support MTL files alongside of OBJ I believe.

Rushmore75:
(glb files support loading pictures directly in the glb file it's self)

PR #99, opened 2026-06-05
Add stl support

Rushmore75:
- [x] STLs load
- [x] clippy is happy

Somewhat related to https://github.com/orhun/ratty/issues/78

Got slightly carried away in `src/inline.rs` and abstracted some logic into a `from` in `src/model.rs`. But besides that this just adds stl support. It looks like `gltf` support basically already exists, but I don't have a gltf model on hand to test it.


src/model.rs
```diff
@@ -297,6 +318,49 @@ fn object_asset_path(path: &Path) -> anyhow::Result<String> {
         .to_string())
 }
 
+fn load_stl_meshes_from_path(path: &Path) -> anyhow::Result<Mesh> {
+    let reader = std::fs::read(path)?;
+    let mut c = Cursor::new(reader);
+    let stl = stl_io::read_stl(&mut c)?;
+
+    // credit: bevy_stl (MIT)
```
orhun, line 392: Why not just use [`bevy_stl`](https://github.com/nilclass/bevy_stl) in this implementation?

orhun (commented):
Looks good! Just had a question about which library to use here.
I saw `bevy_stl` and thought that it might be used since we already use bevy. Additionally, I saw that we already depend on some code from it.

Also, can you do the following updates:

- Update the RGP protocol spec based on this new support. Should be quick to do :)
- Do you have a cool STL model that we can include in `assets/objects` for demo purposes?

.gitignore
```diff
@@ -1,4 +1,5 @@
 /target
 assets/img
+/widget/target
```
orhun: nit: we could just one entry such as `*/**/target` (don't remember the exact syntax) instead of ignoring each target directory

orhun (commented):
Another thing that I realized: I could not use a STL file in place of the cursor. I think this implementation only supports placing STL objects via RGP protocol.

src/model.rs
Rushmore75: `bevy_stl` uses `stl_io` under the hood - So I just took the piece of the (albet, already small) library that this needed. This was because I didn't want to figure out if `bevy_stl` needed any custom loaders or whatnot. https://github.com/nilclass/bevy_stl/blob/73a0afb5d7d4f444a0e5bb95a7f4f83e6bf9f624/src/lib.rs#L61 

This was just the first way I tried to solve it 🤷‍♂️

Rushmore75:
- [x] support stl for cursor
- [x] update RGP protocol spec
- [x] figure out target .gitignore
- [x] find / create demo stl

Rushmore75:
That should solve all those issues if you would like to test again

orhun (commented):
Looks good :)

I tried this STL file for the cursor: https://p.orhun.dev/teamug.stl

But couldn't get it to work properly:

https://github.com/user-attachments/assets/0d112244-98c4-4e54-a69e-41e23287525a

Any ideas?

P.S. can you resolve conflicts? 👼🏼

Rushmore75:
When I open that STL in blender it also has the large plane below it. So I think that it's rendering as-intended 🤔
image

orhun:
Oh I see, I'll try with a different model!
In the meantime, can you run `cargo fmt` and commit the changes?

src/inline.rs
```diff
@@ -545,6 +519,14 @@ pub struct KittyInlineObject {
 
 /// RGP-backed inline object.
 pub enum RgpInlineObject {
+    /// STL mesh payload.
+    Stl {
+        /// The loaded mesh
+        mesh: Mesh,
+        /// This gets created on the fly when it's actually needed.
+        /// If you are creating a RgpInlienObject, chances are that you can set this to `None`.
```
orhun: nit: typo :)

```suggestion
        /// If you are creating a [`RgpInlineObject`], chances are that you can set this to `None`.
```

orhun (commented):
thanks!

I just gave this a try and it works! There is no texture loading yet it seems like, but this should be a good start already :)

PR #107, opened 2026-06-09
Render terminal texture with direct Vello path

### LINKED ISSUE #97: Wonky font scaling
## Summary

Describe the bug clearly and concisely.

## Steps to reproduce

1. Open ratty
2. Look at the font/text
3. See that it's not scaled correctly

## Expected behavior

Correct font alignment and rendering.

## Environment

- OS: Fedora Kinoite 44 (Aurora/Bazzite)
- Ratty version / commit: 0.4.1
- GPU / graphics stack: AMD/Mesa 26.0.8
- Config changes: 
[window]
width = 1440
height = 900
[font]
family = "Rec Mono Duotone"
[cursor.model]
visible = false

## Additional context

Having an issue where the font doesn't render at the correct initial size defined in the config, as well as having a vertical scaling issue causing distortion with Starship and text in general. Using the Ctrl-+ keybind to increase text size can resolve the vertical stretching effect but still throws the size off. Pictured is supposed to be 18pt font by default, but is closer to 13-14:

Image

### LINKED ISSUE #106: weird color support
## Summary

partial problems with color support

## Steps to reproduce

1. open ratty
2. color

## Expected behavior

normal color

## Environment

- OS: mac newest
- Ratty version / commit: ratty newest
- GPU / graphics stack:
- Config changes: font size and type change

## Additional context

Image

Image

Image

### LINKED ISSUE #115: Window can't shrink below 180×120 — Bevy's default WindowResizeConstraints isn't overridden
##Summary

  Allow the Ratty window to shrink below the current ~180×120 logical pxfloor, 
  either by exposing it as config keys (e.g. `[window] min_width` / `min_height`)
  or by lowering/unsetting the default constraint.

## Motivation

 I'm using Ratty as a small always-on-top desktop readout bar on Hyprland
 (launched with `-c` + a custom config), and I can't get the window below
  roughly 180×120 logical px no matter what I try:

  - A compositor resize down to 60px clamps back to 120 tall.
  - Lowering `[window] scale_factor` doesn't change the floor.
  - Disabling the 3D cursor model doesn't change it either.

  So it's a fixed window minimum, independent of scale or content. It blocks
  thin status bars and small widget tiles.

  I think I found the cause: Bevy's default `WindowResizeConstraints` is exactly
  `min_width: 180.0, min_height: 120.0` — a perfect match for the floor I'm
  hitting. A code search of this repo for `resize_constraints` turns up nothing,
  so it looks like Ratty inherits Bevy's default rather than setting its own.

## Proposed approach

Either:

  1. Expose it as config keys (`[window] min_width` / `min_height`), or
  2. Lower / unset the default constraint.

  Something like:

  resize_constraints: WindowResizeConstraints {
      min_width: 1.0,
      min_height: 1.0,
      ..default()
  },

##Additional context

  Totally understand if there's a reason the floor is there. Thanks for making
  Ratty — it's genuinely delightful. 🧀

gold-silver-copper:
## Summary

This PR replaces Ratty's CPU-readback terminal rendering with a **direct Vello→Bevy GPU texture path**: `parley_ratatui` records a Vello scene, the scene is handed to Bevy's render world through a double-buffered exchange, and Vello rasterizes it straight into a Bevy-owned GPU texture with no CPU readback. On top of that core change it reworks the scale/zoom model to render at native display scale, makes the terminal systems `Send` so they run continuously and in parallel, gates per-frame work behind dirty/state checks, and fixes a series of color, scale, shutdown, and GPU-validation bugs.

Updates `parley_ratatui` to `0.3.1` and Bevy to `0.19.0-rc.2`.

Fixes #97.
FIxes #106
Fixes #115 

---

## 1. New direct Vello→Bevy render path (replaces CPU readback)

The old path rendered the terminal on a private offscreen `wgpu` device, read the pixels back to the CPU, and uploaded them into a Bevy `Image` every frame. That whole bridge is gone.

**`src/direct_render.rs` (new file) — `DirectTerminalRenderPlugin`:**
- **`DirectTerminalSceneExchange`** — an `Arc`-backed `Resource` shared between the main world and the `RenderApp`. Its inner state is two single-slot, mutex-guarded mailboxes: `pending: Mutex<Option<DirectTerminalFrame>>` (latest main→render frame, newest wins) and `recycled: Mutex<Option<Scene>>` (one spare Vello `Scene` flowing render→main). `publish_frame` replaces the pending slot and recycles any displaced frame's scene; `take_recycled_scene`/`recycle_scene` (which calls `scene.reset()` and keeps at most one) implement **double-buffered scene ownership** so neither side reallocates a scene per frame.
- **`update_direct_terminal_frame`** (main side) — does the double-buffer dance: take a recycled scene, swap it into the `TerminalRenderer`, build the scene from the Ratatui buffer (`build_scene_with_elapsed`), swap the spare back, and `publish_frame` with both image handles, size, and `base_color`.
- **`extract_terminal_frame`** (`ExtractSchedule`) — moves the pending frame into the render world's `ExtractedDirectTerminalFrame`, recycling any frame it supersedes.
- **`render_terminal_frame`** (RenderGraph, set `RenderGraphSystems::Begin`) — lazily builds the Vello `GpuRenderer`, rasterizes into the render texture, and copies into the present texture. Placed in `Begin` so Vello's submission is ordered **before** the camera passes that sample the terminal texture.
- **Shutdown fix** — the Vello renderer lives in `DirectTerminalRenderState { renderer: SyncCell<Option<GpuRenderer>> }`. `vello::Renderer` is `Send` but not `Sync`, so a plain resource is impossible; the previous non-send resource pinned `render_terminal_frame` to a thread and **deadlocked the pipelined render app's final update during shutdown** (`World::clear_all` waited forever). `SyncCell` makes it a normal `Send` resource.

**`src/terminal.rs`** — deletes the entire `OffscreenGpu` struct (its own `wgpu::Device`/`Queue`, `GpuRenderer`, `TextureTarget`, `TextureReadback`, CPU `rgba: Vec<u8>`), the `pollster::block_on` init, and the `render_to_rgba8…` → `image.data.copy_from_slice` readback. `TerminalSurface` becomes a regular `#[derive(Resource)]`.

**`Cargo.toml`** — removes `pollster` (no CPU readback to block on). **`README.md`** — rewrites the pipeline section: the terminal image is now "fully GPU-resident; the only data crossing from the main world to the render world each frame is the recorded scene, not pixels." **`src/lib.rs`** — adds `mod direct_render;`.

---

## 2. Color correctness & GPU-validation safety (sRGB / storage split)

Vello writes **sRGB-encoded, display-ready** bytes into its `Rgba8Unorm` storage target (confirmed by a readback probe). Two problems had to be solved, and the final design uses **two textures**:

- **`new_terminal_render_image`** — the texture Vello rasterizes into: `Rgba8Unorm`, `STORAGE_BINDING | COPY_SRC | COPY_DST`, **no sRGB view**. Vello binds it as a compute storage target, which requires a plain (non-sRGB) view.
- **`new_terminal_image`** — the **present** texture materials/sprite sample: `Rgba8Unorm`, `TEXTURE_BINDING | COPY_DST` (no storage), with `view_formats = [Rgba8UnormSrgb]` and a `texture_view_descriptor` so it is sampled through an **sRGB view** that decodes the bytes on read (rather than treating them as linear and re-encoding at the swapchain, which washed out colors).

`render_terminal_frame` rasterizes into the render texture, then `copy_texture_to_texture` (a plain same-format `Rgba8Unorm`→`Rgba8Unorm` texel copy) into the present texture each frame.

**Why two textures:** earlier the present texture *was* the storage texture with an sRGB view bolted on. wgpu rejects an sRGB view of a storage texture, so on stricter backends bind-group creation failed (`TextureView with '' label is invalid`) and the app quit. Splitting them keeps the sRGB view on a non-storage texture (Bevy's standard render-target pattern, valid on every backend); the storage texture has no sRGB view. Inspired by `linebender/bevy_vello`, which keeps sRGB off the storage texture.

**Frame retry & zero-fill** — frames that arrive before the GPU image is prepared, or while its size lags a resize, are now **retained and retried** on the next render frame (a newer published frame supersedes a retained one during extraction) instead of being silently dropped, and both textures are zero-filled so a frame sampled before the first copy reads transparent black rather than uninitialized memory.

**`src/rendering.rs`** — `sync_plane_texture` only takes `materials.get_mut(...)` when the `base_color_texture` handle actually differs (checked via an immutable `get`), because `get_mut` marks the material modified and forces a GPU re-prepare.

---

## 3. Scale & zoom model (fixes #97)

Previously cell metrics were floored to whole physical pixels at a forced scale factor of `1.0`, so a font-size step whose advance delta was under one pixel changed only the cell **height** — zoom collapsed into a vertical-only stretch.

- **Native scale by default.** `WindowConfig.scale_factor` (`src/config.rs`) changes from `f32` (default `1.0`) to `Option<f32>` (default `None` = "use the display's scale factor"). `src/main.rs`'s new `window_resolution()` helper applies `with_scale_factor_override(...)` **only** when the config sets it; otherwise the window renders at the display's real scale factor. `config/ratty.toml` comments out `scale_factor` with documentation.
- **Don't double-count the override.** `render_scale_for_window()` (`src/terminal.rs`) chooses its base scale factor as `scale_factor` when a Bevy override is set, else `base_scale_factor()` — "a scale-factor override already defines the physical/logical ratio; applying the backend factor on top of it over-sizes fullscreen terminal textures."
- **Fractional cell quantization.** `build_terminal_renderer` builds the `parley_ratatui` renderer with `CellQuantization::Fractional` and `new_scaled(..., render_scale)`, so both axes grow on every font-size step. A regression test (`font_size_steps_scale_cells_on_both_axes`) asserts `cell_width`/`cell_height` strictly increase across sizes 8..=24 at scales 1.0 and 2.0.
- **Font size in points.** `FontConfig.size` is now interpreted as **points**; the renderer applies `PT_TO_PX = 96.0/72.0` before handing the size to Parley (so the default `18` renders at 18pt). `config/ratty.toml` documents this.
- **`TerminalLayout`** (new) bundles `cols`, `rows`, `texture_size: UVec2` (physical px), `logical_size: Vec2` (Bevy world units), and `render_scale`; `pty_pixels()` clamps to portable-pty's `u16` API. `resize_to_fit(logical_size, render_scale)` recomputes the grid from `renderer.logical_metrics(render_scale)` (logical metrics keep the column/row count stable across HiDPI scales) and returns a layout that the resize/render/input systems all consume. `set_render_scale` rebuilds the renderer; `char_dimensions` now returns fractional `Vec2` logical metrics instead of a ceiled `UVec2`; `snapped_translation()` snaps the sprite to the physical pixel grid (the "align zoom resize layout" fix).

---

## 4. Scene & presentation

**`src/scene/mod.rs`:**
- **Two images in `setup_scene`** — creates `new_terminal_render_image` (storage, `terminal.render_image_handle`) and `new_terminal_image` (present, `terminal.image_handle`); the legacy `create_terminal_image` is kept only for the 3D "back" debug image.
- **Layout-driven sizing** — `setup_scene` becomes a `SetupSceneParams` system param that reads the real `PrimaryWindow` and `TerminalRuntime`, computes `layout = terminal.resize_to_fit(window_size, render_scale_for_window(window))`, resizes the PTY to `layout.pty_pixels()`, and derives all image sizes, the `TerminalViewport`, the sprite `custom_size`/snapped translation, and both plane transform scales from `layout` instead of `app_config.window.{width,height}`.
- **Presentation entities** — `TerminalSprite` (2D, samples the present texture), `TerminalPlane` + `TerminalPlaneBack` (3D meshes, `StandardMaterial { unlit: true, alpha_mode: Blend }`; front samples the present texture, back samples the debug image). Two cameras: `Camera2d` (`order: 0`) and `TerminalPlaneCamera` (`Camera3d`, `order: 1`, orthographic, `clear_color: None` by default).
- **`apply_terminal_presentation` only mutates GPU state when it changes** — front-material `cull_mode` (`None` for Mobius's double-sided ribbon, else `Some(Face::Back)`) is only taken via `get_mut` when it differs; the 2D camera's `is_active = !is_3d` and the 3D camera's `clear_color` (the 3D camera stays active everywhere — it renders the cursor model/RGP objects even in 2D — so "whichever camera renders first owns the screen clear") are only assigned when changed.
- **`sync_terminal_layout`** (new) re-applies a `TerminalLayout` to live entities (viewport, sprite size/translation, plane scales) on resize, backed by new `TerminalSpriteLayoutQuery` / `TerminalPlaneLayoutQuery` / `TerminalPlaneBackLayoutQuery` aliases.

**`src/config.rs`** — adds `TERMINAL_RENDER_TEXTURE_LABEL` for Vello's storage texture and clarifies `TERMINAL_TEXTURE_LABEL` as the sampled present texture.

---

## 5. Scheduling, parallelism & performance

- **`Send` resources.** `src/runtime.rs`: `TerminalRuntime` becomes `#[derive(Resource)]` by wrapping its `!Sync` PTY handles in `SyncCell` (`rx: SyncCell<Receiver<Vec<u8>>>`, `master: SyncCell<Option<Box<dyn MasterPty + Send>>>`) and exposing `try_recv(&mut self)`. With `TerminalRuntime` and `TerminalSurface` both `Send`, `src/main.rs` inserts them with `insert_resource` (was `insert_non_send_resource`) and every consuming system in `src/systems.rs` flips `NonSend*` → `Res*`, so Bevy can schedule them off the main thread / in parallel. `pump_pty_output` now calls `runtime.try_recv()`.
- **Continuous updates.** `src/main.rs` replaces the per-focus reactive throttle (`UpdateMode::reactive_low_power(FOCUSED_UPDATE_INTERVAL)`) with `WinitSettings::continuous()` — Bevy's default switches unfocused windows to reactive mode, which would delay background PTY output. The `Instant`-based redraw throttle in `TerminalRedrawState` is removed.
- **`TerminalRedrawSet` pipeline.** The old `redraw_soft_terminal` god-system (which redrew, uploaded, refreshed the back texture, synced materials, *and* drove cursor-model loading) is split into three `.chain()`-ed systems in a new `TerminalRedrawSet`: `render_terminal_widget` → `sync_terminal_materials` → `finish_terminal_model_load`, ordered `.after()` input/resize/PTY systems.
- **Frame-dirty + blink-tick gating.** `render_terminal_widget` computes a shared `TerminalFrameDirty` from `needs_redraw || blink_ticked || !model_loaded`, where `blink_ticked` uses a `Local<u64>` bucket over `BLINK_TICK_SECS = 0.25` (rapid blink half-period; slow blink is a multiple). Texture content only changes with terminal state or blink phase (warp/camera motion is mesh-/camera-side), so an idle terminal re-rasterizes only ~4×/s instead of every frame, and the downstream pipeline systems early-return on clean frames. `sync_image` likewise only takes `get_mut` on each image when its size changes.
- **Run conditions** (`src/plugin.rs`) so continuous mode doesn't burn work: `apply_terminal_presentation` runs only when presentation/plane-view/Mobius state `is_changed()`; `apply_inline_objects` only when presentation changed or inline-object entities were `Added`; `sync_rgp_objects` only when RGP objects exist; `animate_mobius_transition` only in Mobius mode or during a transition; `animate_terminal_plane_warp` only when not `Flat2d`; `sync_asset_to_terminal_cursor` only when the cursor model is visible.
- `src/systems.rs` `handle_window_resize` is simplified to call `terminal.resize_to_fit(...)` + `sync_terminal_layout(...)` instead of inlining grid/viewport math, and the redraw pipeline (not the resize system) uploads the resized image.

---

## 6. Input handling under the new layout

- **`src/keyboard.rs`** — font-zoom bindings now resize through the window-based layout model: instead of computing the grid from the viewport and integer char dimensions, it calls `terminal.resize_to_fit(window_size, render_scale_for_window(window))`, resizes the PTY from `layout.pty_pixels()`, and propagates geometry with `sync_terminal_layout(...)`. This keeps zoom proportional under fractional cell metrics. Params switch `NonSendMut` → `ResMut` and gain the window + layout queries.
- **`src/mouse.rs`** — mouse-to-cell mapping is corrected for a viewport that is smaller than and centered in the window (letterbox/pillarbox under the new scale model). `position_to_cell` gains a `window_size` parameter, subtracts the half-margin (`(window_size - viewport.size) * 0.5`), returns `None` when the cursor is outside the terminal area, and then clamps within the viewport — previously it clamped raw window coordinates, assuming the viewport filled the window from the origin. The wheel `Pixel` branch drops a now-unneeded `.max(1) as f32` since `char_dimensions()` returns fractional `Vec2`.

---

## 7. Dependency & Bevy 0.19 migration

- **`Cargo.toml`** — `bevy` `0.18.1` → `0.19.0-rc.2`; adds the `bevy_scene` feature; `parley_ratatui` `0.2.1` → `0.3.1`; removes `pollster`.
- **`parley_ratatui` 0.3.1** also fixes terminal-cell rendering upstream: it snaps cell background fills to the pixel grid (no base-color bleed between cells with fractional cell sizes) and renders block elements such as the half block (`▀`) as exact geometric fills instead of font glyphs (which only filled the em box and left gaps in the taller line-height cell). The present texture keeps **linear** filtering because it is mapped onto transformable/scaled 3D geometry where point sampling would pixelate it; the seam fixes are made at authoring time in `parley_ratatui`, not by the sampler.
- **Scene type rename** — Bevy 0.19 renamed `Scene`→`WorldAsset` / `SceneRoot`→`WorldAssetRoot`: `src/inline.rs` (`Handle<Scene>` → `Handle<WorldAsset>`) and `src/model.rs` (`SceneRoot` → `WorldAssetRoot`) are mechanical migrations.
- **Other API renames** — `insert_non_send_resource`/`init_non_send_resource` → `insert_non_send`/`init_non_send` (`AppWindowIcon`, `TerminalClipboard`); `StandardMaterial.shadows_enabled` → `shadow_maps_enabled`; `apply_plane_warp` takes `Option<AssetMut<'_, Mesh>>`.

---

## Bug fixes in this PR (by commit)

- **font zoom proportional with fractional cell metrics** (#97) — vertical-only zoom stretch.
- **avoid double-counting fullscreen scale** — explicit scale override was multiplied by the backend factor, oversizing fullscreen textures.
- **align zoom resize layout** — sprite/plane sizing and pixel-snapped translation on resize.
- **unblock shutdown by making render state a Send resource** — non-send Vello renderer deadlocked shutdown.
- **decode terminal texture as sRGB and retry dropped frames** — washed-out colors and stale/uninitialized frames.
- **give Vello a storage texture and copy into an sRGB present texture** — `TextureView … is invalid` crash on strict backends (sRGB view on a storage texture).
- **update parley_ratatui to 0.3.1** — inter-cell seams and block-glyph gaps.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1140,7 +1140,7 @@ dependencies = [
  "wasm-bindgen",
  "wasm-bindgen-futures",
  "web-time",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2178,7 +2178,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2313,7 +2313,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3629,7 +3629,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -4049,7 +4049,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d8fae84b431384b68627d0f9b3b1245fcf9f46f6c0e3dc902e9dce64edd1967"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -4149,9 +4149,9 @@ dependencies = [
 
 [[package]]
 name = "parley_ratatui"
-version = "0.3.0"
+version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "40efa408c041aad816471c16eb2b1be909fed2ddb789b6b46099e9f0549c2f2a"
+checksum = "87e3cc2adfe7f5622f17d285f144ef2f4d7c096f9bcd0bbe0a13bad4283b60e3"
 dependencies = [
  "naga",
  "parley",
@@ -4892,7 +4892,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.12.1",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -6513,7 +6513,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1140,7 +1140,7 @@ dependencies = [
  "wasm-bindgen",
  "wasm-bindgen-futures",
  "web-time",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2179,7 +2179,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2314,7 +2314,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -3630,7 +3630,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -4050,7 +4050,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d8fae84b431384b68627d0f9b3b1245fcf9f46f6c0e3dc902e9dce64edd1967"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -4150,9 +4150,9 @@ dependencies = [
 
 [[package]]
 name = "parley_ratatui"
-version = "0.3.2"
+version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a3c4a0467a959680e789f88eb5b69afc8970a39b1a591c7b29dc15076dc97406"
+checksum = "993017a77e34db00adbbf64e51db1bf17ff4cfa3273e93901cf16e90713f14e4"
 dependencies = [
  "naga",
  "parley",
@@ -4578,14 +4578,15 @@ checksum = "ca45419789ae5a7899559e9512e58ca889e41f04f1f2445e9f4b290ceccd1d08"
 
 [[package]]
 name = "ratatui"
-version = "0.30.1"
+version = "0.30.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1695748e3a735b34968c887ceea5a380b43545903868ae8f5b666593100f6b68"
+checksum = "3274ba0a2c5e1bcad2a2005d20f4dc59dad26b2eb0940fb094500dba4099d57d"
 dependencies = [
  "instability",
  "ratatui-core",
  "ratatui-crossterm",
  "ratatui-macros",
+ "ratatui-termina",
  "ratatui-termwiz",
  "ratatui-widgets",
  "serde",
@@ -4593,15 +4594,14 @@ dependencies = [
 
 [[package]]
 name = "ratatui-core"
-version = "0.1.1"
+version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "42d3603f354bba8c595fa47860e60142d7372b7210c27044c6a7d0e1a4336b44"
+checksum = "cbb175c433c8e28a809d1f5773a2ae96e68c0ce40db865cbab1020bf33ae479c"
 dependencies = [
  "bitflags 2.13.0",
  "compact_str",
  "critical-section",
  "hashbrown 0.17.0",
- "indoc",
  "itertools",
  "kasuari",
  "lru",
@@ -4616,9 +4616,9 @@ dependencies = [
 
 [[package]]
 name = "ratatui-crossterm"
-version = "0.1.1"
+version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2b2867bedcbd6a690ca4f8672a687b730ec07660c79844517b084311b529980c"
+checksum = "567584a3b0e6a8203c23de40b4861497266725eb5363dbfd18a1edd603cca9f0"
 dependencies = [
  "cfg-if",
  "crossterm",
@@ -4628,19 +4628,30 @@ dependencies = [
 
 [[package]]
 name = "ratatui-macros"
-version = "0.7.1"
+version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "80fac59720679490d89d200df411faa249be728681adcabed3d047ae72c48f1d"
+checksum = "ed7dc68daa7498a43e4d68e0eb078427e10c38fbcfbb1e42d955f1fa2140d814"
 dependencies = [
  "ratatui-core",
  "ratatui-widgets",
 ]
 
 [[package]]
-name = "ratatui-termwiz"
-version = "0.1.1"
+name = "ratatui-termina"
+version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "386b8ff8f74ed749509391c56d549761a2fcdb408e1f42e467286bcb7dac8967"
+checksum = "c0bf912d9e66f057a759d92e386a280ea886b352ab757d6ac4d653c7ed2c43c2"
+dependencies = [
+ "instability",
+ "ratatui-core",
+ "termina",
+]
+
+[[package]]
+name = "ratatui-termwiz"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "faf03e0380b7744054d6cb74224fe3adf062a029754933f575ca1e3b4c2ce977"
 dependencies = [
  "ratatui-core",
  "termwiz",
@@ -4648,9 +4659,9 @@ dependencies = [
 
 [[package]]
 name = "ratatui-widgets"
-version = "0.3.1"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7ef4f17dd7ac3abf5adc2b920a03c61eee4bfe6a88fa5191936895525371d79c"
+checksum = "66e3d19bcc9130ca376277d93b60767ff121ace3be06f5f95f81dd68956407d1"
 dependencies = [
  "bitflags 2.13.0",
  "hashbrown 0.17.0",
@@ -4893,7 +4904,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.12.1",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -5376,6 +5387,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "06794f8f6c5c898b3275aebefa6b8a1cb24cd2c6c79397ab15774837a0bc5755"
 dependencies = [
  "winapi-util",
+]
+
+[[package]]
+name = "termina"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9048a889effe34a5cddee0af7f53285198b16dca3be510858d38dfdb3e62a04e"
+dependencies = [
+ "bitflags 2.13.0",
+ "parking_lot",
+ "rustix 1.1.4",
+ "signal-hook",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -6514,7 +6538,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1140,7 +1140,7 @@ dependencies = [
  "wasm-bindgen",
  "wasm-bindgen-futures",
  "web-time",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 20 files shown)


orhun:
Can you also update the [rendering pipeline](https://github.com/orhun/ratty#rendering-pipeline) docs? I think it will help with understanding this PR better as well.

gold-silver-copper:
will do

orhun:
Getting this while trying to run:

```
2026-06-14T22:48:36.481488Z ERROR bevy_render::error_handler: Caught rendering error: Validation Error

Caused by:
  In Device::create_bind_group, label = 'StandardMaterial'
    TextureView with '' label is invalid

2026-06-14T22:48:36.481532Z ERROR bevy_render::error_handler: Caught rendering error: Validation Error

Caused by:
  In Device::create_bind_group, label = 'sprite_material_bind_group'
    TextureView with '' label is invalid

2026-06-14T22:48:36.482422Z  INFO vello::wgpu_engine: Initialising shader modules in parallel using 14 threads
2026-06-14T22:48:36.643692Z ERROR bevy_render::error_handler: Quitting the application due to Validation RenderError
2026-06-14T22:48:36.702724Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702749Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702763Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702768Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702774Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702778Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702782Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702793Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702802Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702807Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702814Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702828Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702832Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702837Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702842Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702848Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702854Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702860Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702866Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702869Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702874Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702880Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702886Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702912Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702920Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702930Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702935Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702941Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
2026-06-14T22:48:36.702950Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
```

gold-silver-copper:
@orhun 

The error you got should be fixed, was a Vulkan vs Metal wgpu issue. Could you try rerunning to test, i dont have a decent linux machine here for it

gold-silver-copper:
PR is ready for review here

orhun:
it all works fine!

orhun (approved):
yessir

PR #108, opened 2026-06-09
Shut down gracefully on window close

gold-silver-copper:
## Summary

- emit AppExit directly when the primary window requests close
- shut down the PTY runtime when Bevy exits
- make PTY child/master/writer cleanup explicit and idempotent
- keep Drop as a backup shutdown path for TerminalRuntime


orhun (approved):
thx!

PR #117, opened 2026-06-23
add gpuBackend/gpuAdapter module options for wgpu device selection under Nix

DarthPJB:
The NixOS and Home Manager modules now expose two options for controlling wgpu device selection:
programs.ratty = {
  gpuBackend = "vulkan";    # "vulkan" | "gl" | "gles" | null
  gpuAdapter = "RTX 3060";  # substring match | null
};
gpuBackend — Sets WGPU_BACKEND to force the graphics API. Cases:
- "vulkan" — Explicit Vulkan path. Needed when the default backend auto-detection picks a broken backend, or when testing driver-specific behaviour.
- "gl" / "gles" — OpenGL fallback. Useful on systems where Vulkan is unavailable or broken (headless, VNC, older hardware, driver bugs).
- null — Auto-detect (default). wgpu picks the best available backend.
gpuAdapter — Sets WGPU_ADAPTER_NAME to select a specific GPU by name substring. Cases:
- Multi-GPU systems (integrated + discrete, or dual discrete) where wgpu picks the wrong adapter.
- Testing against a specific GPU without disabling others in BIOS.
- Offloading rendering to a secondary GPU while keeping the primary for display.
The NixOS module composes these into a single makeWrapper call alongside the existing --config-file flag. The Home Manager module sets them via home.sessionVariables.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -55,6 +55,30 @@
                 }
               '';
             };
+
+            gpuBackend = lib.mkOption {
+              type = lib.types.nullOr (lib.types.enum [ "vulkan" "gl" "gles" ]);
+              default = null;
+              description = ''
+                Force the wgpu backend.
+                Set to null (default) for auto-detection.
+                Useful when the Vulkan ICD loader selects an incompatible driver
+                or when running in headless/VNC environments where OpenGL is preferred.
+              '';
+              example = "vulkan";
+            };
+
+            gpuAdapter = lib.mkOption {
+              type = lib.types.nullOr lib.types.str;
+              default = null;
+              description = ''
+                Substring match to select a specific GPU adapter name.
+                Set to null (default) to use the system default.
+                Useful on multi-GPU systems (e.g. integrated + discrete)
+                or when wgpu picks the wrong adapter.
+              '';
+              example = "RTX 3060";
+            };
           };
         };
     in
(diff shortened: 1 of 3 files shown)


DarthPJB:
@orhun - I am unsure if this is the best way to pass in an environment variable like this; I am not 'rust-native' - Please feel free to amend the approach or suggest a better methodology.

orhun (approved):
thank you!

Nomagno:
@DarthPJB @orhun FYI, on my system (Xubuntu 24.04 with proprietary nvidia drivers and a GTX 1060 6GB) this change causes ratty to straight up crash:

```
2026-07-05T22:37:50.696170Z ERROR wgpu::backend::wgpu_core: Handling wgpu errors as fatal by default

thread 'Compute Task Pool (0)' (45707) panicked at /home/nomagno/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-27.0.1/src/backend/wgpu_core.rs:1588:26:
wgpu error: Validation Error

Caused by:
  In Device::create_texture
    Texture format Rgba16Unorm can't be used due to missing features
      Features Features { features_wgpu: FeaturesWGPU(TEXTURE_FORMAT_16BIT_NORM), features_webgpu: FeaturesWebGPU(0x0) } are required but not enabled on the device


note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in system `<Enable the debug feature to see the name>`!
```

Nomagno:
If I revert the changes to `main.rs`, then it doesn't crash anymore :P

orhun:
Ah damn, that's not good! Any chance that you could experiment with the other settings while setting up the Bevy app? I wonder if removing that line is a good solution. 

cc @Nomagno @DarthPJB

DarthPJB:
@orhun I just saw this, apologies for the late reply.

I replicated the crash, so I'm trying a few changes locally; I'm quite a slow-boat here.

I'll try and provide a better implementation soon.

DarthPJB:
I certainly have no objection to reverting the line in question as a short term solution.

orhun:
Let's revert it :) Can you drop another PR?

Issue #114, opened 2026-06-17
Add camera controls to the Ratty Graphics Protocol


Nomagno:
I have implemented my own take on this this in 5652d631c94faa6feb04a193b579b95c11f69947

Specifically, see the changes to [`protocols/graphics.md`](https://github.com/orhun/ratty/blob/5652d631c94faa6feb04a193b579b95c11f69947/protocols/graphics.md) in said commit (there's a new 'Camera' section at the end of the file).

Basically, a new `c` command is added, which allows to control the camera mode (flat/orthogonal/perspective #89/mobius) of the camera, and in the case of the 3D modes its zoom level, position, and rotation. There are 10 camera setting slots that can be preconfigured, and a flag that can be used to switch to the configured slot immediately. The idea is that applications can preconfigure slots for simplicity and then just switch between them, and Ctrl+Alt+[0-9] could even be used by the user in the future to switch between the slots configured by the application, such as for different camera modes in a 3D viewer/videogame.

My only worry with making a PR is that, for my implementation to allow the user controls and the commands to change the camera at the same time with no overwriting, I have had to basically hard-couple the order of most of the rendering and I/O pipeline. This might not be acceptable for performance/architecture. Any feedback is appreciated. I can whip up test code for my commit if needed

orhun:
Hey, I think this is super interesting to have! Feel free to drop a PR :)

Just as a reference, updating the [big rat](https://github.com/orhun/ratty/blob/main/widget/examples/big_rat.rs) demo would be great for showcasing this :)

Nomagno:
Sorry, I meant to close the other PR not this issue 💀

PR #118, opened 2026-06-27
Ratty camera protocol

Nomagno:
Hey!
Ctrl + Alt + O -> Regular orthographic view
Ctrl + Alt + P -> New perspective view ('videogame camera' style)
Basically, both cameras are created at startup, but only the requested one is activated in Bevy. Not sure if this might suppose a performance penalty. I am also not an expert in Rust nor Bevy, so sorry if something's improvable :sweat_smile:. I also refactored a bunch of `match` statements into comparisons against `is_3d()` where I thought it might be reasonable, since 3 long enum values per arm seemed silly once I added the new `Persp3d` mode (I can revert this if needed).

Additionally:

The ratty graphics protocol is expanded as per the update to `protocols/graphics.md`, with the corresponding implementation. The rendering and input processing pipeline is reordered in order to allow the applications and the user to control the camera at once harmoniously. I am not a Rust expert, and would, in general, appreciate feedback on the architecture and style.

Missing: better documentation of the code, add support for the features described here to the big rat demo, and perhaps add `Ctrl`+`Alt`+`[0-9]` bindings for manually changing to the 10 camera slots.

Obsoletes #89 and implements #114


src/inline.rs
```diff
@@ -242,10 +255,47 @@ impl TerminalInlineObjects {
...
+                        camera_slots.slots[camera_slot as usize].0.camera_offset[1] = py;
+                    }
+                    if let Some(pz) = settings.offset[2] {
+                        camera_slots.slots[camera_slot as usize].0.camera_offset[2] = pz;
+                    }
+                    if let Some(rx) = settings.rotation[0] {
+                        camera_slots.slots[camera_slot as usize].0.yaw = rx;
+                    }
+                    if let Some(_ry) = settings.rotation[1] {
+                        /*camera_slots.slots[camera_slot as usize].0.roll = ry;*/
```
orhun: Why is this disabled?

src/inline.rs
```diff
@@ -242,10 +255,47 @@ impl TerminalInlineObjects {
...
+    fn handle_rgp_sequence(&mut self,
+                           sequence: &[u8],
+                           camera_slots: &mut TerminalCameraViewSlots) -> Option<Option<Vec<u8>>> {
         let operation = consume_rgp_sequence(sequence)?;
         Some(match operation {
             RgpOperation::SupportQuery => Some(support_reply()),
+            RgpOperation::Camera {
+                camera_slot, switch_immediately, settings
+            } => {
+                //info!("camera settings for slot {}: set {}; rotation: ({},{},{})", camera_slot, switch_immediately, settings.rotation[0].unwrap_or(0.0), settings.rotation[1].unwrap_or(0.0), settings.rotation[2].unwrap_or(0.0));
```
orhun: Please remove the commented out code :)

orhun (commented):
thanks for the PR! I have pushed a cleanup commit and have a small question about one of the parts :)

src/inline.rs
Nomagno: Ah, my bad, should've mentioned this as an open implementation question! It's because ratty's camera doesn't currently have the capability to be rolled by the user, it has 2 axis rotation. If I allowed to actually change the roll, the user would have no intuitive way to correct it afterwards with the mouse 😅
There's several solutions, ignoring roll inputs from the Ratty Camera Protocol is the one that seemed safest for a first implementation.

orhun:
Hey hey! I just fixed conflicts and gave this a try - but I'm not sure if it works as expected.
When I switch to perspective mode, the terminal disappears. But maybe that's expected? I'm not sure.

Do you mind giving this a try with the latest changes here?

Nomagno:
Sure! Can you confirm if it worked before the merge conflict resolution? Just to have another data point

orhun:
Hmm maybe I broke it 😅

Nomagno:
Yep, I can confirm latest commit is broken for me too, but it works before the changes

orhun:
okay I just ran this from your previous commits and got an idea how the feature should look like. I think something broke while fixing the conflicts, any chance that you could take a look or reimplement this?

Nomagno:
Sure, I'll have a look!

Nomagno:
Yeah uhhhh, I'm having trouble merging in a bug-free manner too :sob: hard to believe the branch could've gone stale this quickly, I'll do a more in depth investigation

orhun:
Yeah... it's because of our refactor in #107 probably.

Maybe @gold-silver-copper can also take a look here

gold-silver-copper:
I'll take a look when I get the chance

Nomagno:
Closing since my changes were already merged. Thank you!
In the future, feel free to ping me, as I thought this was still on my backburner of tasks for the past month :sweat_smile:

I'll review that my tests still work, and look into expanding the existing demos.
Also, yes, I *was* wondering about the perspective inversion bug I possibly caused but forgot to double check it, thanks for fixing that!

gold-silver-copper:
Oops sorry! I forgot to comment here. Yes it was merged as a stacked PR on top of ur branch

PR #119, opened 2026-06-30
feat(cursor): apply texture from config for obj and stl models

onuralpszr:
This adds a `texture` option to the cursor model config so a base-color texture can be applied to the cursor model without editing code. Addresses #78.

```toml
[cursor.model]
path = "CairoSpinyMouse.obj"
texture = "texture.png"
```

The path is resolved relative to the config file, the same way `path` is.

- `config`: new `texture: Option<PathBuf>` field on `[cursor.model]`, resolved relative to the config file.
- `model`: decode the image and set it as `base_color_texture` on the cursor material. Also load UV coordinates from OBJ files, which were previously dropped (a texture has nothing to map to without UVs).
- `systems`: thread the image asset store into cursor spawning.
- `config/ratty.toml`: documented example.

### DIFF
--- a/config/ratty.toml
+++ b/config/ratty.toml
@@ -23,6 +23,7 @@ size = 18
 
 [cursor.model]
 path = "CairoSpinyMouse.obj"
+# texture = "texture.png"
 scale_factor = 6.0
 brightness = 0.5
 x_offset = 0.5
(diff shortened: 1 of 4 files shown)


onuralpszr:
@orhun friendly ping* also I tested it was able to handle nicely stl/obj + texture you apply (of course proper wrap type picture would make better look but that's using blender etc. type of skills required)

orhun (approved):
Nice, thanks!

PR #121, opened 2026-07-02
perf: reduce inline object memory churn in 3D warp mode

### LINKED ISSUE #120: perf: inline object memory churn in 3D warp mode
## Problem

In 3D presentation with warp active, `sync_inline_objects` runs every frame via `force_warp_sync`, despawned all inline entities, and respawned meshes/materials without cleaning up Bevy asset stores. Kitty images also kept a duplicate RGBA copy in the inline registry after GPU upload.

This causes unnecessary RAM growth during 3D + warp sessions with inline objects (RGP / Kitty).

## Proposed fix

- Remove per-frame inline rebuild in warp mode; animate Kitty plane mesh vertices in place
- Reuse cached Kitty plane mesh/material handles and remove stale assets on layout changes
- Move Kitty RGBA into Bevy `Image` on upload instead of cloning
- Reuse STL extruded mesh handles keyed by depth; remove superseded OBJ/STL assets on depth change

## Context

Follow-up to the architectural note in the README about CPU readback being a separate concern. This targets inline-object asset lifecycle, which is a more immediate win.

Local Windows benchmarks showed ~100 MB lower steady-state working set under 3D + warp + inline RGP object (~962 MB → ~863 MB), with idle 2D unchanged.

jackthepunished:
Fixes #120

## Summary

- Stop rebuilding all inline scene entities every frame in 3D warp mode; animate Kitty image planes in place instead (same approach as `animate_terminal_plane_warp` for the terminal surface)
- Reuse cached Kitty plane mesh/material handles across syncs and remove stale assets when layout subdivisions change
- Move Kitty RGBA bytes into the Bevy `Image` on upload instead of keeping a duplicate copy in the inline registry
- Reuse STL extruded mesh handles keyed by depth and remove superseded OBJ/STL mesh assets when extrusion depth changes

## Problem

In 3D presentation with warp active, `sync_inline_objects` ran every frame via `force_warp_sync`, despawned all inline entities, and respawned meshes/materials without cleaning up Bevy asset stores. Kitty images also kept a full RGBA copy after GPU upload.

## Approach

- Remove `force_warp_sync`; inline sync now runs only when `TerminalInlineObjects::needs_sync()` is true
- Add `animate_inline_kitty_planes` to update Kitty plane vertex positions in place during warp animation
- Cache plane assets on `KittyInlineObject` and drop registry-side RGBA after first upload
- Align STL mesh caching with the existing OBJ depth-key pattern

## Benchmarks

Measured locally on Windows, release build.

**Idle 2D (~30s):** no meaningful change (~833 MB working set before and after)

**3D + warp + inline RGP object (~3 min, sampled every 15s):**

| | Before | After |
|--|--------|-------|
| Steady-state working set | ~962 MB | ~863 MB |
| Growth over 3 min | +7.6 MB, then flat | +2.1 MB, flat |

## Test plan

- [x] `cargo check`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [x] Manual smoke test: 2D terminal, 3D toggle, warp, inline RGP object
- [ ] Kitty inline images in 3D warp mode
- [x] macOS / Linux sanity check

### DIFF
--- a/src/inline.rs
+++ b/src/inline.rs
@@ -27,6 +27,35 @@ pub struct TerminalInlineObjectSprite;
 #[derive(Component)]
 pub struct TerminalInlineObjectPlane;
 
+/// Layout data used to animate Kitty image planes on the warped terminal surface.
+#[derive(Component, Clone, Copy)]
+pub(crate) struct InlineKittyPlaneLayout {
+    /// Normalized horizontal center within the terminal plane.
+    pub local_x: f32,
+    /// Normalized vertical center within the terminal plane.
+    pub local_y: f32,
+    /// Normalized width within the terminal plane.
+    pub local_width: f32,
+    /// Normalized height within the terminal plane.
+    pub local_height: f32,
+    /// Horizontal mesh subdivision count.
+    pub x_segments: u32,
+    /// Vertical mesh subdivision count.
+    pub y_segments: u32,
+}
+
+/// Cached GPU assets for a Kitty image plane attached to the terminal surface.
+pub(crate) struct KittyPlaneCache {
+    /// Cached horizontal mesh subdivision count.
+    pub x_segments: u32,
+    /// Cached vertical mesh subdivision count.
+    pub y_segments: u32,
+    /// Cached plane mesh handle.
+    pub mesh: Handle<Mesh>,
+    /// Cached plane material handle.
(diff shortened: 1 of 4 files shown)


orhun (approved):
thanks!

PR #123, opened 2026-07-05
fix(input): respect bracketed paste mode and harden paste handling

### LINKED ISSUE #122: Paste sends bracketed paste markers even when the application never enabled them
## Summary

Pasting with <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>V</kbd> always wraps the clipboard text in bracketed paste markers (`ESC[200~` / `ESC[201~`), even when the foreground application never enabled bracketed paste mode (`DECSET 2004`). Applications that don't speak the protocol receive the markers as literal input.

I think I found the cause: the markers are written unconditionally in the paste branch of `handle_keyboard_input` ([src/keyboard.rs#L457-L463 at `8d1c0a5`](https://github.com/orhun/ratty/blob/8d1c0a5/src/keyboard.rs#L457-L463)). The parser already tracks the mode — `vt100::Screen::bracketed_paste()` — but it is never consulted.

There is also a security side to this: the pasted payload is forwarded verbatim, so clipboard text that itself contains `ESC[201~` terminates the bracket early and everything after it is treated as typed input (including a `\r` executing it). This is the paste injection vector that xterm patched as [CVE-2015-8971](https://nvd.nist.gov/vuln/detail/CVE-2015-8971); Alacritty, kitty and wezterm all sanitize bracketed payloads for the same reason.

## Steps to reproduce

1. Run `ratty`
2. Run `cat > /tmp/out` (cat never enables bracketed paste)
3. Copy `hello`, paste with <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>V</kbd>, press <kbd>Ctrl</kbd>+<kbd>D</kbd>
4. `xxd /tmp/out`

## Expected behavior

The file contains only `hello`. Instead it starts with `1b 5b 32 30 30 7e` (`ESC[200~`) and ends with `ESC[201~`.

## Environment

- OS: reproducible on all platforms (the paste path is platform-independent)
- Ratty version / commit: v0.5.0, `8d1c0a5`
- Config: defaults

## Additional context

While in there I noticed non-bracketed pastes also normalize line endings to `\n`, but the Enter key sends `\r` — terminals conventionally deliver pasted line breaks as carriage returns when bracketed paste is off (Alacritty, kitty, wezterm and foot all do this).

I have a fix ready with unit tests (gate the markers on `bracketed_paste()`, strip ESC bytes from bracketed payloads, send `\r` for line breaks otherwise) and will open a PR.

jackthepunished:
## Summary

Fixes #122.

- keyboard: only wrap pasted text in `ESC[200~`/`ESC[201~` when the application has bracketed paste (DECSET 2004) enabled, via `vt100::Screen::bracketed_paste()`
- keyboard: drop the 7-bit `ESC` and 8-bit `CSI` (`U+009B`) control introducers from bracketed payloads so clipboard content containing `ESC[201~` (or its C1 equivalent) cannot terminate the bracket early and inject input
- keyboard: deliver newlines as carriage returns in non-bracketed pastes, matching the byte the Enter key sends

The paste encoding is extracted into `encode_paste`, a pure function, so the behavior is unit-testable. Control-introducer filtering happens per `char` rather than per byte, so it can't corrupt a multi-byte UTF-8 sequence whose continuation byte happens to be `0x9b`.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo test --lib` (5 unit tests: marker gating, newline handling in both modes, and the ESC/CSI injection cases)

To verify manually: run `cat > /tmp/out`, paste, then `xxd /tmp/out` — the file should contain only the pasted text, no `ESC[200~`/`ESC[201~`; pasting at a bash/zsh prompt (which enables bracketed paste) should still arrive as a single bracketed paste.

### DIFF
--- a/src/keyboard.rs
+++ b/src/keyboard.rs
@@ -456,11 +456,8 @@ pub fn handle_keyboard_input(
                 }
                 BindingAction::Paste => {
                     if let Some(text) = params.clipboard.paste() {
-                        let normalized = text.replace("\r\n", "\n").replace('\r', "\n");
-                        let mut bytes = Vec::from(b"\x1b[200~".as_slice());
-                        bytes.extend_from_slice(normalized.as_bytes());
-                        bytes.extend_from_slice(b"\x1b[201~");
-                        params.runtime.write_input(&bytes);
+                        let bracketed = params.runtime.parser.screen().bracketed_paste();
+                        params.runtime.write_input(&encode_paste(&text, bracketed));
                     } else {
                         warn!("failed to read clipboard contents for paste");
                     }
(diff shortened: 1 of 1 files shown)


orhun (approved):
LGTM, thanks!

PR #125, opened 2026-07-07
Scope RGP style updates to the object they target

Tnsr-Q:
## Problem

A single RGP `u;color=` / `brightness=` / `depth=` update on **one** inline object despawned and respawned **every** inline object — sprites, Kitty images, and all RGP 3D objects — because `TerminalInlineObjects.dirty` was one container-level bool consumed by `sync_inline_objects`, which drains all three entity queries. Every `RgpAnimationState` accumulator reset, so every object on screen visibly jumped poses. (`tools/silk/docs/rgp-truths.md` under-reported this as respawning "the whole object", singular.)

## Change

Invalidation now has three granularities:

1. **`color`/`brightness` on OBJ/STL objects → in-place material rewrite.** A new `apply_rgp_restyle` system recomputes the object's material from its anchor style (shared `rgp_object_material` + `apply_brightness` helpers, extracted from the previously duplicated spawn blocks) and writes it through `Assets::<StandardMaterial>::get_mut`. No despawn, no accumulator reset, pose stays continuous. Rewritten entities are marked `BrightnessAdjusted` so `apply_instance_brightness` never multiplies brightness on top.
2. **`depth` → per-object rebuild.** Depth re-extrudes the mesh, so a rebuild is legitimate — but now only that object despawns/respawns (`rebuild_objects: HashSet<u32>`). `color`/`brightness` on **glTF** objects take this path too: their materials come from the scene asset, so style cannot rewrite them in place.
3. **Viewport/grid changes and everything else → the existing global rebuild.** Updates on ids without an RGP entity mapping (Kitty images, unregistered ids) also keep the conservative global path — their entities carry no object id, so a scoped despawn cannot find them.

A full sync subsumes and clears any queued per-object work (`finish_sync`).

## Not changed (flagged for a protocol-level decision)

`silk`'s `TWEENABLE_FIELDS` still rejects `color`/`brightness` in tweens. Now that mesh-backed restyles are live in ratty, the ban *could* be lifted — but older terminals still respawn (scene-wide!) on every such update, glTF objects still rebuild, and the RGP capability string has no key advertising live restyle for authors to gate on. Lifting the ban would make casts that play smoothly on new ratty and stutter badly on old ratty. Decide separately; the doc comment on `TWEENABLE_FIELDS` explains the state.

## Docs

- `tools/silk/docs/rgp-truths.md`: "Updates: live vs respawn" rewritten for the new granularity (and the old scene-wide blast radius recorded); animation-accumulator note updated.
- `protocols/graphics.md`: update-verb semantics reworded — authors should still treat `depth`/`color`/`brightness` as scene-setup fields, not streaming fields.
- **Not updated (follow-up):** the rgp-composer skill docs (`.claude/skills/rgp-composer/SKILL.md` hard-rules bullet and `references/rgp.md` "respawn split" section) still describe the old always-respawn mechanism. The authoring rules they enforce remain correct (never tween those fields), so nothing they teach produces a bad cast — but the mechanism description should be refreshed to match `rgp-truths.md` in a separate change.

## Tests

- `src/inline.rs`: routing tests updated intentionally — depth rebuilds only its object, mesh color/brightness restyle in place, depth supersedes a pending restyle, glTF routes to rebuild, unmapped ids keep the global path, `finish_sync` clears queued granular work.
- Wire format untouched: `cargo test` in `tools/silk` (golden transmissions byte-identical) and in the repo root both pass, including the locked v1 bit-identical animation tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Tnsr-Q:
Apologies — opened against the wrong base repo; this change belongs on the fork.

PR #128, opened 2026-07-13
Added mouse_scroll_lines option

veksha:
Now mouse wheel scrolls screen by 3 lines (configurable)

### DIFF
--- a/config/ratty.toml
+++ b/config/ratty.toml
@@ -8,6 +8,7 @@ opacity = 0.8
 default_cols = 104
 default_rows = 32
 scrollback = 2000
+mouse_scroll_lines = 3
 
 # [shell]
 # program = "/bin/bash" 
--- a/src/config.rs
+++ b/src/config.rs
@@ -159,6 +159,8 @@ pub struct TerminalConfig {
     pub default_rows: u16,
     /// Scrollback line count.
     pub scrollback: usize,
+    /// Number of lines scrolled per mouse wheel tick in flat 2D mode.
+    pub mouse_scroll_lines: usize,
 }
 
 impl Default for TerminalConfig {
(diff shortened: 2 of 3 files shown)


orhun:
Hey, this looks good! I'm wondering when/why this is useful, could you maybe explain it?

veksha:
Hi. when you want to scroll up with your mouse long wall of text it helps to scroll a little faster.

config/ratty.toml
```diff
@@ -8,6 +8,7 @@ opacity = 0.8
 default_cols = 104
 default_rows = 32
 scrollback = 2000
+mouse_scroll_lines = 3
```
orhun, line 11: Maybe we can rename this to `mouse_scrollback`? Just to align with the `scrollback` option above.

orhun (approved):

config/ratty.toml
veksha: or scrollback_speed, or something like that. choose what you wish :)

orhun: I asked some LLM:

>   Examples: Alacritty uses scrolling.multiplier, kitty uses wheel_scroll_multiplier, Ghostty uses mouse-scroll-multiplier, WezTerm uses scrollback_lines for retained history.
>
> • For this PR, I’d keep mouse_scroll_lines.
>
> It’s better than both options you were considering:
>
> - mouse_scrollback sounds like the size of the scrollback buffer, especially since ratty already has scrollback = 2000.
> - mouse_speed is too broad and could mean cursor movement, sensitivity, selection speed, drag speed, etc.
> - mouse_scroll_lines says exactly what the PR does: number of lines scrolled per mouse wheel tick.
>
> If you want to align more with other terminal naming, mouse_scroll_multiplier is also defensible, but since the value is literally line count and the PR says “scrolls screen by 3 lines,”
mouse_scroll_lines is the clearest name for this implementation.

so yeah, I think it's all good to go :)

PR #131, opened 2026-07-15
feat(config): add window update_mode option

### LINKED ISSUE #130: Expose winit `UpdateMode` as config — `WinitSettings::continuous()` costs ~10x idle CPU for small always-on windows
## Summary

Ratty 0.5.0 sets `WinitSettings::continuous()` unconditionally. For small, always-on "ambient" Ratty windows (a status bar / readout that repaints every few seconds), this burns **~53% of a CPU core per window at idle**, versus **~5.5%** on 0.4.2 — roughly 10x more, continuously.

I'd like a way to opt a window into a lower-power update mode via config, with `continuous()` staying the default.

## Motivation

I run two small Ratty windows as a permanent desktop bar: a 520x60 readout and a 90x60 widget. They're ambient — the readout only repaints every 3 seconds. On 0.5.0 the pair costs **over a full core, 24/7**, which makes them impractical to leave running.

I isolated the cause. Building **v0.5.0 with only the `WinitSettings` resource reverted** to 0.4.2's values, changing nothing else:

| Build | Idle CPU per window |
|---|---|
| v0.5.0 (stock) | ~53% |
| v0.5.0, **only** `WinitSettings` reverted | **~7-8%** |
| v0.4.2 | ~5.5% |

So it's specifically the `WinitSettings` change, not the Vello render rewrite that landed in the same release.

I also wanted to flag that the stated reason for `continuous()` looks orthogonal to the cost. The comment reads:

```rust
// Always update continuously, focused or not. Bevy's default switches
// unfocused windows to a reactive mode, which would delay background PTY output.
```

But 0.4.2 already guarded that case explicitly, with `unfocused_mode: UpdateMode::Continuous`:

```rust
// 0.4.2
.insert_resource(WinitSettings {
    focused_mode:   UpdateMode::reactive_low_power(FOCUSED_UPDATE_INTERVAL), // 33ms
    unfocused_mode: UpdateMode::Continuous,
})
```

Background PTY output was already continuous there. The 10x cost comes from the **`focused_mode`** change (33ms-capped reactive -> unbounded continuous), so keeping `unfocused_mode: Continuous` preserves the anti-PTY-delay behaviour regardless of what `focused_mode` does.

I'm assuming the unbounded `focused_mode` may well be deliberate — smooth 3D/RGP at high refresh — which is why I'm asking for a knob rather than a revert.

## Proposed approach

A config key in the `[window]` section, defaulting to current behaviour:

```toml
[window]
# "continuous" (default) | "low_power"
update_mode = "low_power"
frame_interval_ms = 33   # only used by low_power
```

roughly mapping to:

```rust
match app_config.window.update_mode {
    UpdateMode::Continuous => WinitSettings::continuous(),
    UpdateMode::LowPower => WinitSettings {
        focused_mode: UpdateMode::reactive_low_power(interval),
        unfocused_mode: UpdateMode::Continuous, // keep PTY output prompt
    },
}
```

Default unchanged, so nothing shifts for existing users. Either that or any other shape you prefer — exposing it at all is the ask.

## Additional context

- Measured with `top`, settled idle, same machine / config / window sizes across all three builds. Happy to test any change against this setup.
- Environment: Arch Linux, Hyprland 0.55.4 (Wayland), NVIDIA RTX 4080 (driver 610), Ratty built from source.
- Related: #115 — my earlier request for a lower window minimum, which is what makes windows this small possible in the first place.
- These windows are Hyprland `no_focus` overlays, so in practice they appear to sit on `focused_mode`.

carltenzler-ctrl:
Closes #130.

Adds `[window] update_mode` (`"Continuous"` | `"LowPower"`) and `[window] frame_interval_ms`, exposing the winit `UpdateMode` used while the window is focused.

**The default is unchanged.** `update_mode` defaults to `Continuous`, which takes the same `WinitSettings::continuous()` path as today, so existing setups behave exactly as they do now.

`LowPower` restores what 0.4.2 used: `focused_mode: reactive_low_power(frame_interval_ms)` with `unfocused_mode: Continuous`. Keeping unfocused continuous preserves the reasoning behind the current code — the comment notes that a reactive unfocused mode would delay background PTY output, and that rationale is about the unfocused mode, so `LowPower` only changes the focused one. `frame_interval_ms` defaults to `33`, matching 0.4.2's `FOCUSED_UPDATE_INTERVAL`.

Measured on the setup from #130 (RTX 4080, Hyprland, 520x60 window, idle, same build — only the config differs):

| Config | CPU, one core |
|---|---|
| default (`Continuous`) | 49.3% |
| `update_mode = "LowPower"` | 7.85% |

The checks in CONTRIBUTING.md pass: `cargo check`, `cargo fmt --all -- --check`, and `cargo clippy --all-targets --all-features -- -D warnings`.

Two notes:

- I used PascalCase values (`"Continuous"` / `"LowPower"`) to match the existing enum values in `ratty.toml` (`style = "Regular"`, `action = "Copy"`) rather than the lowercase spelling I used in the issue. Easy to change if you'd prefer.
- No tests added — there's no existing test setup for config parsing, and the default path is unchanged. Happy to add some if you want them.

I'm not a Rust developer, and I used Claude to help write this patch. I've built and tested it on the machine from #130, and I'm glad to test any changes you'd like against that setup.

### DIFF
--- a/config/ratty.toml
+++ b/config/ratty.toml
@@ -3,8 +3,8 @@ width = 960
 height = 620
 opacity = 0.8
 # scale_factor = 1.0
-# update_mode = "Continuous"  # or "LowPower" to redraw reactively while focused
-# frame_interval_ms = 33      # frame rate cap in "LowPower" mode
+# update_mode = "Continuous"  # Use "LowPower" to redraw only after input or terminal output.
+# frame_interval_ms = 33      # Minimum milliseconds between LowPower redraws.
 
 [terminal]
 default_cols = 104
--- a/config/ratty.toml
+++ b/config/ratty.toml
@@ -3,6 +3,8 @@ width = 960
 height = 620
 opacity = 0.8
 # scale_factor = 1.0
+# update_mode = "Continuous"  # or "LowPower" to redraw reactively while focused
+# frame_interval_ms = 33      # frame rate cap in "LowPower" mode
 
 [terminal]
 default_cols = 104
--- a/src/config.rs
+++ b/src/config.rs
@@ -136,6 +136,11 @@ pub struct WindowConfig {
     pub scale_factor: Option<f32>,
     /// Window opacity from `0.0` to `1.0`.
     pub opacity: f32,
+    /// How the window schedules redraws while focused.
+    pub update_mode: UpdateModeConfig,
+    /// Minimum time between redraws while focused, in milliseconds.
+    /// Only applies when `update_mode` is `LowPower`.
+    pub frame_interval_ms: u64,
 }
 
 impl Default for WindowConfig {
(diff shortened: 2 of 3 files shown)


orhun (approved):
thanks!

PR #138, opened 2026-07-28
Swap the vt100 terminal engine for rio-vt

raphamorim:
Replaces the [vt100](https://crates.io/crates/vt100) parser with [rio-vt](https://crates.io/crates/rio-vt), Rio's terminal engine, as the VT state machine behind ratty. The PTY layer (`portable-pty`) is unchanged, so cross-platform support (incl. `x86_64-pc-windows-msvc`) is preserved — this is purely an engine swap.

To keep the diff contained, rio-vt is wrapped in a small vt100-shaped shim (`src/vtshim.rs`: `Parser` / `Screen` / `Cell` / `Color` / `MouseProtocolMode` + a `Listener` implementing rio-vt's `EventListener`), so the render/input call sites change mostly by import path. Cell colours/attributes are resolved from rio-vt's packed `Square` + per-grid style table; DA/DSR/cursor-report replies now come from rio-vt itself (`RioEvent::PtyWrite`).

## Why rio-vt

rio-vt is a safe-Rust, dependency-light terminal core that unlocks capabilities beyond vt100 for a GPU terminal like ratty:

- **Inline image protocols** — sixel, Kitty graphics, iTerm2 (decoded by the core, surfaced via events).
- **Shell integration** — OSC 7 (cwd), OSC 8 (hyperlinks), OSC 133 (semantic prompt / command boundaries).
- **Correct Unicode width & grapheme clustering**, regex search over screen + scrollback, per-row damage tracking.
- Actively maintained and in production at [Lovable](https://lovable.dev). Also drops the transitive **`net2` (RUSTSEC-2020-0016, unmaintained)** vt100 pulled in.

## Status

`cargo build`, `cargo test` (9 pass), and `cargo clippy` are green; the `ratatui-ratty` widget crate still builds. Since I can't run the GPU UI here, a few things want a visual pass: cell colours/attributes (bold/dim/italic/underline/inverse, 256/truecolor), cursor visibility, mouse-protocol mode detection, scrollback, and resize reflow (now rio-vt's native behaviour). The primary DA reply now comes from rio-vt (VT220-class) rather than a fixed `\x1b[?1;2c` — functionally equivalent for capability detection, exact bytes differ.

A couple of rio-vt gaps I worked around and would upstream: no public accessor for the current kitty-keyboard flags (reconstructed from `Mode` bits) and no `modifyOtherKeys` modelling.

- Benchmark vs vt100 & alacritty_terminal: https://github.com/raphamorim/rio-vt-benchmark
- Engine background: https://rioterm.com/blog/2026/07/27/rio-vt-and-librio

**Cursor pending-wrap**: rio-vt reports col `width-1` at end-of-line (with a pending-wrap flag) vs vt100's `width` — same alacritty-style semantics as everywhere; at most a 1-cell cursor-at-line-end visual.


orhun:
It works generally fine! Needs a conflict resolution though

gold-silver-copper:
it's a good idea I think, since rio-vt is more maintained, and has more features, it makes sense to move to it.

orhun:
Have you experienced any weirdness in runtime? @gold-silver-copper

raphamorim:
lemme know anything and i can update the pr as well.  need to fix the conflicts

raphamorim:
i've added the vt shim but i can refactor that too, would break the api

raphamorim:
fixed conflicts

PR #142, opened 2026-08-10
Support the kitty image protocol natively through rio-vt

raphamorim:
Now that the vt100 engine is gone (#140), we can lean on rio-vt for the kitty graphics protocol too. This deletes ratty's hand-rolled APC parser and enables rio-vt's `graphics` feature instead, so the whole protocol — chunked transmissions, PNG/RGB/RGBA payloads, placement ids, deletes, queries, quiet modes, the 320MB eviction budget, and per-screen state on alt-screen swaps — is handled by the engine, which already answers acks and queries through the same reply path we use for DA/DSR.

On ratty's side, kitty state is now derived from the engine each frame: decoded pixels arrive through `RioEvent::UpdateGraphics` and become Bevy textures, and placements are resolved from the engine's absolute, scrollback-aware positions. That fixes a few things the old parser couldn't do:

- images scroll with the content they were placed next to, including into and out of history (the old row-text diffing heuristic is gone)
- multiple placements of the same image share one texture instead of re-decoding
- source crops (`x=`/`y=`/`w=`/`h=`) and sub-cell offsets (`X=`/`Y=`) are honored, both for sprites and for the warped 3D planes
- kitty z-index ordering is respected between images
- `kitten icat --unicode-placeholder` style virtual placements are matched against the engine's placement registry rather than guessed from cell colors alone

The PTY grid resize now passes pixel dimensions through to the engine, since placements are sized against the cell pixel metrics. Sixel stays unadvertised in DA1: the engine can decode it now, but ratty only renders kitty placements.

Tests cover direct and chunked transfers, RGB expansion, crops, scrollback tracking, deletes dropping their textures, placeholder-driven placements, and queries being acked without placing anything.

kitty-ratty

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1818,6 +1818,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2ec7c5eb7a16992b1904d76c517d170ab353b0e0b3d5a0c81a8a0cd1037893cf"
 
 [[package]]
+name = "color_quant"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3d7b894f5411737b7867f4827955924d7c254fc9f4d91a6aad6b097804b1018b"
+
+[[package]]
 name = "colorchoice"
 version = "1.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2485,6 +2491,15 @@ checksum = "77ce24cb58228fbb8aa041425bb1050850ac19177686ea6e0f41a70416f56fdb"
 
 [[package]]
 name = "font-types"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "39a654f404bbcbd48ea58c617c2993ee91d1cb63727a37bf2323a4edeed1b8c5"
+dependencies = [
+ "bytemuck",
+]
+
+[[package]]
+name = "font-types"
 version = "0.11.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5b38ad915f6dadd993ced50848a8291a543bd41ca62bc10740d5e64e2ab4cfd7"
@@ -2506,7 +2521,7 @@ dependencies = [
  "objc2-core-text",
  "objc2-foundation 0.3.2",
  "parlance",
- "read-fonts",
+ "read-fonts 0.39.2",
  "roxmltree",
  "smallvec",
  "windows",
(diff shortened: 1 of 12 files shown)


raphamorim:
Ran a double review over this branch (a general correctness pass with adversarial verification, plus an independent kitty-conformance/integration pass) and pushed fixes for everything that survived:

- **textures now come straight from the engine's image store**, validated by transmission timestamp. The event-queue copy could desync when an alt-screen app reused an image id from the main screen, or when eviction dropped the inactive screen's copy of a shared id — both now just rebuild from whatever the engine holds.
- **direct placements use the engine's `kitty_overlay_geometry` + `clip_overlay_to_rect`** instead of a hand-copied version of the same math. That deletes the duplication and fixes clipping: a placement scrolled half off the top now shows the bottom slice of the image instead of overhanging the grid.
- **virtual (U=1) placements render per placeholder row-run** using the engine's `IncompletePlacement` decoding and `compute_run_geometry`, so the row/column diacritics, indexed-color image ids (`38;5;n`), the placement id in the underline color, and the id high byte all work now. A placeholder image half-scrolled into history shows the correct half instead of squashing the whole image into the visible rows.
- **negative z-indices layer behind the terminal surface** per the spec (sprites below the text quad, planes biased behind the surface), and 3D planes get a per-placement depth bias so overlapping images stack deterministically instead of z-fighting.
- **resizes pass the renderer's fractional cell metrics** to the engine instead of `pw / cols`, whose truncation mis-scaled native-size images by up to ~6% with fractional font metrics. Also covers the font-size-change resize path.
- **the per-frame refresh is gated** on the engine's dirty flag, the pending-redraw flag, and a scroll-state snapshot, so idle frames (and all frames without graphics) cost a few comparisons; `refresh_kitty_graphics` is now explicitly ordered after window resizes.
- **APC scanning only intercepts RGP sequences** and streams everything else to the engine as it arrives — a multi-megabyte single-APC kitty transfer is no longer buffered and rescanned per chunk, and a stray unterminated APC introducer can't freeze the display.

Three engine-side issues surfaced that belong in rio-vt rather than here, filing separately: `a=p` for a nonexistent image and every `a=q` reply OK unconditionally where kitty would report ENOENT; deletes never remove virtual (U=1) placements; and atlas (sixel/iTerm2) pixel data is tracked against the 320MB budget with no untrack path once the update queues are drained.

Now 116 tests, including per-run slicing, indexed-color placeholders, alt-screen id collisions, clipped scrolling, z ordering, and refresh gating.

raphamorim:
The three engine-side issues from the review are fixed upstream and released in rio-vt 0.5.20 (raphamorim/rio@390a66f477, 531b897293, e6772529be), so this PR now depends on it:

- `a=p` referencing an image the engine doesn't hold answers `ENOENT` instead of `OK`, and `a=q` runs the real decode path before replying — clients get honest feature detection and retransmit evicted images. There's a ratty-side test pinning the ENOENT reply end-to-end through the PTY reply path.
- deletes (`d=a`/`d=i`/`d=n`) now remove virtual (U=1) placements, so placeholder cells stop rendering deleted images.
- atlas (sixel/iTerm2) graphics give their bytes back to the 320MB budget when released, so heavy `imgcat` use no longer starves kitty images into permanent eviction.

No ratty code changes were needed for the bump — 117 tests pass against 0.5.20.

orhun:
I just gave this a try, with the document demo available in the repo

The image looks like this (templeos logo)

image

With some block background, is that expected? @raphamorim

gold-silver-copper:
Closing because ratty now uses its own ratty-vt

PR #135, opened 2026-07-19
Fix/nix module improvements

DarthPJB:
This is my current WIP  for correcting my prior work, as well as improving the module usage for nix users overall.

My aim here is to provide declarative configurable GPU  usage, as well as some nice-to-haves building on the prior work.

PR #144, opened 2026-08-29
Fix crash caused by #135

Nomagno:
This is only a temporary fix, as discussed in the comments of #135

This code was causing the render plugin to crash if it would not find a WebGPU driver, or something to that effect (further investigation required):

```rs
...
.set(RenderPlugin {
    render_creation: bevy::render::settings::RenderCreation::Automatic(Box::new(
        WgpuSettings {
            priority: WgpuSettingsPriority::WebGPU,
            ..default()
        },
    )),
    ..default()
})
...
```

The error itself:
```
2026-08-29T19:00:04.762939Z ERROR bevy_render::error_handler: Caught rendering error: Validation Error

Caused by:
  In Device::create_texture
    Texture format Rgba16Unorm can't be used due to missing features
      Features Features { features_wgpu: FeaturesWGPU(TEXTURE_FORMAT_16BIT_NORM), features_webgpu: FeaturesWebGPU(0x0) } are required but not enabled on the device
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,8 +2,6 @@
 use anyhow::anyhow;
 use bevy::asset::AssetPlugin;
 use bevy::prelude::*;
-use bevy::render::RenderPlugin;
-use bevy::render::settings::{WgpuSettings, WgpuSettingsPriority};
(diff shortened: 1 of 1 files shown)


orhun (approved):
yolo

PR #154, opened 2026-09-09
Add Rust performance tooling, migrate Python automation, and optimize ASCII parsing

gold-silver-copper:
## Summary

Adds reproducible performance tooling for Ratty, replaces the repository's Python automation with Rust, lands one measured VT engine optimization, and forks `vte` into `ratty-vte` so UTF-8 chunk-boundary bugs are fixed in the parser rather than worked around.

### `ratty-vte`: vte fork with the UTF-8 boundary fixes applied directly

New workspace crate `crates/ratty-vte`, forked from vte 0.15.0. Apache-2.0 OR MIT licensing, upstream authors, and provenance are retained; the optional `ansi` module and `serde` feature are dropped because ratty-vt only uses `Parser`, `Perform`, and `Params`. Two chunk-dependent parser bugs are fixed in `advance_partial_utf8`:

- Completing a UTF-8 scalar split across `advance` calls now consumes only that character's bytes, so the character after it is no longer dropped (alacritty/vte#159).
- A UTF-8 encoded C1 control split across `advance` calls is executed, matching unsplit input (alacritty/vte#156).

Five new tests cover the upstream reproducers, byte-wise versus whole-buffer dispatch parity for mixed text, all 32 C1 controls, and invalid partial scalars not consuming their terminating byte.

ratty-vt now depends on ratty-vte. Its byte-at-a-time drain around partial scalars is removed and `process` is a plain `advance` again; its exhaustive chunk-cut regression tests remain and pass against the fork. ratty-vt is bumped to 0.1.1 because its contents diverge from the published 0.1.0. The publish workflow accepts `ratty-vte`; the documented order is ratty-vte, ratty-vt, ratty.

### Measured optimization (ratty-vt)

Printable-ASCII fast path in grapheme clustering: when both the previous cell's final byte and the incoming character are printable ASCII, boundary analysis is skipped. Non-ASCII input keeps the existing Unicode path.

| Workload | Configuration | Metric | Baseline | Candidate | Change | Samples |
| --- | --- | --- | ---: | ---: | ---: | --- |
| ASCII processing | 120×40, 2,000 history rows, 4,096-byte chunks, 4 MiB/sample | Median of five run medians, ms | 131.162 | 83.454 | −36.37% | 5 alternating pairs × 7 samples; per-pair reductions 34.98–38.41% |

The 93 workloads shared by the baseline and candidate matrices showed no median regression above 5%. Baseline source is `05308cd` plus the harness. Apple M2 Max, optimized CPU-only runs. Raw JSON, the comparison summary, and the CPU sample are retained locally and can be attached on request.

### Python → Rust migration

| Removed | Replacement |
| --- | --- |
| `.github/scripts/headless-smoke.py` (orchestration, config generation, PNG/log validation, results) | `cargo run --locked -p xtask -- smoke --output <dir>` |
| Embedded styled-text PTY producer | `xtask emit text` |
| Embedded short-lived scene producer | `xtask emit eof` |

All seven smoke cases, diagnostics, input-delivery checks, timeouts, kill-and-reap, and exit statuses are preserved; PNG validation now decodes the image. Parity was established before deletion: reports matched, six PNGs were byte-identical, and the document capture differed only in animated mesh silhouettes. No maintained automation invokes Python any more.

### Tooling added (unpublished `xtask` workspace member)

- `bench-vt`: VT matrix over 3 grids, 5 chunk sizes, ASCII/Unicode/styles/cursor/alt-screen/scroll-region/reflow with state validation outside the timed region.
- `bench-app`: CPU drawing and PTY replay through the real runtime and `drain_pty_output`, with byte-count, digest, final-state, and child-exit checks.
- `bench-interactive` / `bench-flood`: burst input and cursor-reply validation, sustained-output acknowledgement latency proxy.
- Desktop workloads and an opt-in `performance` feature recorder (`RATTY_PERF_REPORT`) for real-window runs, graphics lifecycle, and font/DPI/window transitions. Normal builds contain none of it.
- `profiling` Cargo profile (release optimizations with symbols).

Run `cargo run -p xtask -- --help` for the available commands; each subcommand documents its inputs, validation, and what its metric does not establish.

### What is not claimed

- No application, desktop, GPU, or input-to-display gain is claimed. Controlled application comparisons were repeatedly contaminated by external CPU load on the measurement machine and remain outstanding; the existing runs are retained as correctness evidence only.
- GPU timestamps are unavailable through the pinned renderer on Metal.
- CI exercises the harness for correctness with no timing thresholds.

## Verification

- Root: fmt, check, test (all-features and default), doctests, clippy (`--all-features -D warnings`), docs (`-D warnings`), and `cargo package --workspace --exclude xtask` verifying ratty-vte, ratty-vt, and ratty together.
- Widget: fmt, check, test, clippy.
- The full 96-workload VT matrix validates against ratty-vte, including the three fragmented Unicode cases that fail on stock vte 0.15.0.
- Independent reviews of the full diff, the review fix-up, and the fork commit; all findings applied (shared `PTY_QUEUE_ACCOUNTING_BOUND` constant, CI all-features clippy plus a default-feature test run, perf-build exit-code ordering, executable check parity in the smoke runner, fork MSRV and C1 documentation).


orhun:
Ayo, what's the status on this 👀

